PR #8, opened 2017-01-31
rename SmallInt to Int in example

FlyingFoX:
this way the definition above of the enum MyIntEnum is sufficient to
understand the example code.


JelteF:
Good spot, I'm in the process of rewriting the documentation (see the doc directory), so I might remove it in the near future again.

PR #28, opened 2017-12-11
Remove double different

dvberkel:
different was written twice in a row. One should suffice


JelteF:
Thanks, changed the same doc in `lib.rs` myself.

PR #44, opened 2018-03-21
Fix leftovers from the copy from constructor.rs in into.rs

### LINKED ISSUE #40: Weirdness about #[derive(Into)]
When using `#[derive(Into)]`, I got an error saying `Only structs can derive a constructor`.

Looking at the code, I see it was copy/pasted from `constructor.rs` and still talks about `#[derive(Constructor)]`, and the error message is still that of `constructor.rs`. I'm not sure why Into is not supported for things other than structs...

And the other suprise I got reading the source is that `#[derive(Into)]` actually implements... `From`. Which is understandable in some way, since `From` implies `Into`, but then why doesn't it support the same things as `#[derive(From)]`?

glandium:
Fixes #40

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,7 +8,7 @@ repository = "https://github.com/JelteF/derive_more"
 documentation = "https://jeltef.github.io/derive_more/derive_more/"
 
 readme = "README.md"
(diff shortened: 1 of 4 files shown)


JelteF:
Thanks, looks good

PR #46, opened 2018-03-21
Add support for #[derive(TryInto)]

### LINKED ISSUE #45: Add derive(TryInto) for enums
I would be very interested in having TryInfo/TryFrom to convert enum variants into their underlying type.

That is:
```rust

enum Foo {
    A(usize),
    B(String),
}

impl ::std::convert::TryFrom<Foo> for usize {
    type Error = ();
    fn try_from(value: Foo) -> Result<Self, Self::Error> {
        match value {
            Foo::A(v) => Ok(v),
            _ => Err(()),
        }
    }
}

impl ::std::convert::TryFrom<Foo> for String {
    type Error = ();
    fn try_from(value: Foo) -> Result<Self, Self::Error> {
        match value {
            Foo::B(v) => Ok(v),
            _ => Err(()),
        }
    }
}
```

glandium:
Fixes #45

### DIFF
--- a/doc/try_into.md
+++ b/doc/try_into.md
@@ -1,0 +1,157 @@
+% What #[derive(TryInto)] generates
+
+This derive allows you to convert enum variants into their corresponding
+variant types.
+One thing to note is that this derive doesn't actually generate an
+implementation for the `TryInto` trait.
+Instead it derives `TryFrom` for each variant in the enum and thus has an
+indirect implementation of `TryInto` as recommended by the
+[docs](https://doc.rust-lang.org/core/convert/trait.TryInto.html).
+
+As of writing the `TryInto` and `TryFrom` traits are nightly only.
+
+# Example usage
+
+```rust
+#![feature(try_from)]
+# #[macro_use] extern crate derive_more;
+use std::convert::TryFrom;
+#[derive(TryInto)]
+enum MixedData {
+    Int(u32),
+    String(String),
+}
+
+fn main() {
+    let s = MixedData::String("foo".into());
+    assert_eq!("FOO", String::try_from(s).unwrap().to_uppercase());
+}
+```
+
+# Structs
+
+Deriving `TryInto` for structs is not supported because there is no failing
+mode. Use `#[derive(Into)]` instead. `TryInto` will automatically get a
+blanket implementation through `TryFrom`, automatically derived from `From`,
+which `#[derive(Into)]` produces.
+
+# Enums
+
+When deriving `TryInto` for an enum, each enum variant gets its own
+`TryFrom` implementation.
+For instance, when deriving `TryInto` for an enum link this:
+
+```rust
+#![feature(try_from)]
+# #[macro_use] extern crate derive_more;
+# fn main(){}
+#[derive(TryInto)]
+enum MixedInts {
+    SmallInt(i32),
+    BigInt(i64),
+    TwoSmallInts(i32, i32),
+    NamedSmallInts { x: i64, y: i64 },
+    UnsignedOne(u32),
+    UnsignedTwo(u32),
+}
+```
+
+Code like this will be generated:
+
+```rust
+# #![feature(try_from)]
+# enum MixedInts {
+#     SmallInt(i32),
+#     BigInt(i64),
+#     TwoSmallInts(i32, i32),
+#     NamedSmallInts { x: i64, y: i64 },
+#     UnsignedOne(u32),
+#     UnsignedTwo(u32),
+# }
+impl ::std::convert::TryFrom<MixedInts> for (i32) {
+    type Error = &'static str;
+    fn try_from(value: MixedInts) -> Result<Self, Self::Error> {
+        match value {
+            MixedInts::SmallInt(__0) => Ok(__0),
+            _ => Err("Only SmallInt can be converted to i32"),
+        }
+    }
+}
+impl ::std::convert::TryFrom<MixedInts> for (i64) {
+    type Error = &'static str;
+    fn try_from(value: MixedInts) -> Result<Self, Self::Error> {
+        match value {
+            MixedInts::BigInt(__0) => Ok(__0),
+            _ => Err("Only BigInt can be converted to i64"),
+        }
+    }
+}
+impl ::std::convert::TryFrom<MixedInts> for (i32, i32) {
+    type Error = &'static str;
+    fn try_from(value: MixedInts) -> Result<Self, Self::Error> {
+        match value {
+            MixedInts::TwoSmallInts(__0, __1) => Ok((__0, __1)),
+            _ => Err("Only TwoSmallInts can be converted to (i32, i32)"),
+        }
+    }
+}
+impl ::std::convert::TryFrom<MixedInts> for (i64, i64) {
+    type Error = &'static str;
+    fn try_from(value: MixedInts) -> Result<Self, Self::Error> {
+        match value {
+            MixedInts::NamedSmallInts { x: __0, y: __1 } => Ok((__0, __1)),
+            _ => Err("Only NamedSmallInts can be converted to (i64, i64)"),
+        }
+    }
+}
+impl ::std::convert::TryFrom<MixedInts> for (u32) {
+    type Error = &'static str;
+    fn try_from(value: MixedInts) -> Result<Self, Self::Error> {
+        match value {
+            MixedInts::UnsignedOne(__0) | MixedInts::UnsignedTwo(__0) => Ok(__0),
+            _ => Err("Only UnsignedOne, UnsignedTwo can be converted to u32"),
+        }
+    }
+}
+```
+
+When deriving `TryInto` for an enum with Unit variants like this:
+
+```rust
+#![feature(try_from)]
+# #[macro_use] extern crate derive_more;
+# fn main(){}
+#[derive(TryInto)]
+enum EnumWithUnit {
+    SmallInt(i32),
+    Unit,
+}
+```
+
+Code like this will be generated:
+
+```rust
+# #![feature(try_from)]
+# enum EnumWithUnit {
+#     SmallInt(i32),
+#     Unit,
+# }
+impl ::std::convert::TryFrom<EnumWithUnit> for (i32) {
+    type Error = &'static str;
+    fn try_from(value: EnumWithUnit) -> Result<Self, Self::Error> {
+        match value {
+            EnumWithUnit::SmallInt(__0) => Ok(__0),
+            _ => Err("Only SmallInt can be converted to i32"),
+        }
+    }
+}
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -111,6 +111,7 @@ before.
 10. [`#[derive(AddAssign)]`](https://jeltef.github.io/derive_more/derive_more/add_assign.html)
 11. [`#[derive(MulAssign)]`](https://jeltef.github.io/derive_more/derive_more/mul_assign.html)
 12. [`#[derive(Constructor)]`](https://jeltef.github.io/derive_more/derive_more/constructor.html)
+13. [`#[derive(TryInto)]`](https://jeltef.github.io/derive_more/derive_more/try_into.html)
```
JelteF: Can you change this in the `src/lib.rs` file instead? I like to update the README only when I create a release.

JelteF: Also can you add a link to the "conversion traits" section above?

JelteF:
I made two small comments already, but I will review this fully in a couple of days.

glandium:
Note the docs tests are broken, and they obviously aren't going to be happy about `#![features]`. A trick would be to run the `external-doc` command with the `RUSTC_BOOTSTRAP=1` environment variable set.

glandium:
A random idea for extending this in the future: an attribute could be added that tells what error type to use instead of `()`, and which would be expected to be `From<TheEnumType>`.

JelteF:
If you change `.travis.yml` to use `DOCS=true` on the nightly build instead of the stable it should probably work.

doc/try_into.md
```diff
@@ -0,0 +1,132 @@
+% What #[derive(TryInto)] generates
+
+This derive allows you to convert enum variants into their corresponding
+variant types.
+One thing to note is that this derive doesn't actually generate an
+implementation for the `TryInto` trait.
+Instead it derives `TryFrom` for each variant in the enum and thus has an
+indirect implementation of `TryInto` as recommended by the
+[docs](https://doc.rust-lang.org/core/convert/trait.TryInto.html).
+
```
JelteF: Can you add a small example here like done in `Display`/`Index`/`IndexMut`.

tests/try_into.rs
```diff
@@ -0,0 +1,89 @@
...
+    NamedBigInts { x: i64, y: i64 },
+    Unsigned(u32),
+    NamedUnsigned { x: u32 },
+    Unit,
+}
+
+#[test]
+fn test_try_into() {
+    let i = MixedInts::SmallInt(42);
+    assert_eq!(TryInto::<i32>::try_into(i), Ok(42i32));
```
JelteF: Is this really the way this trait should be used? Can't you do something like:
```rust
assert_eq!(Ok(42i32), i.try_into())
```

tests/try_into.rs
```diff
@@ -0,0 +1,89 @@
+#![allow(dead_code)]
+
+// TryInto is still an unstable trait. In the meanwhile, use a trick
+// to define the trait (as well as TryFrom) without relying on
+// #![feature(try_from)].
+#![no_std]
+extern crate std as ruststd;
+mod std {
+    pub mod convert {
```
JelteF: I think it's nicer to just not run these tests on stable for now by using this solution to that: https://stackoverflow.com/a/37680212/2570866

doc/try_into.md
```diff
@@ -0,0 +1,132 @@
...
+    fn try_from(value: MixedInts) -> Result<Self, Self::Error> {
+        match value {
+            MixedInts::NamedSmallInts { x: __0, y: __1 } => Ok((__0, __1)),
+            _ => Err(()),
+        }
+    }
+}
+```
+
+When multiple variants have the same type, no `TryInto` is derived for them.
```
JelteF: Maybe I'm missing something, but I think it actually makes sense to derive `TryInto` for variants with the same type as well. Because both can be converted into their inner type. This wasn't the case for `From` because the other way around you don't know which enum variant to choose.

tests/try_into.rs
glandium: You can, but you can't do the same for the Err(()) cases.

doc/try_into.md
glandium: In that example, there would be two `TryFrom<MixedInts> for u32`, that wouldn't work out.

glandium:
Note, I'm starting to wonder if the default Error shouldn't be Self rather than (). What do you think?

doc/try_into.md
JelteF: You could have one `TryFrom<MixedInts> for u32` where multiple match arms return an `Ok(....)` right?

glandium: Oh, yes, indeed.

JelteF:
Thanks for all the work on this, it's looking really nice overall, I see you've already incorporated some of my feedback.

Regarding the Error type, I don't think it's needed to go crazy with configuration options for the first version of this. But I think it's nice to have a descriptive error by default though. For the `Add` derive I've used `Result<T, 'static str>` in cases where enum variants couldn't be added together: https://github.com/JelteF/derive_more/blob/master/src/add_like.rs#L146. I think that should also work here. We could have an error message like `format!("Enum variant {} can't be converted to {}", variantname, type)` (if you think of something clearer/easier to implement that's also fine)

glandium:
> `format!("Enum variant {} can't be converted to {}", variantname, type)`

I'm not sure adding a quadratic number of strings is all that great.

JelteF:
That's a very good point. Probably better without the variant info then, so
it stays linear. Or something like: only {} can be converted to {}

glandium:
```diff
@@ -61,17 +61,33 @@ fn enum_try_into(input: &DeriveInput, data_enum: &DataEnum) -> Tokens {
             quote!((#(#vars),*))
         };
 
+        let output_type = if original_types.len() == 1 {
+            format!("{}", quote!(#(#original_types)*))
+        } else {
+            let types = original_types
+                .iter()
+                .map(|t| format!("{}", quote!(#t)))
+                .collect::<Vec<_>>();
+            format!("({})", types.join(", "))
+        };
+        let variants = variants
+            .iter()
+            .map(|v| format!("{}", v.ident))
+            .collect::<Vec<_>>()
+            .join(", ");
+        let message = format!("Only {} can be converted to {}", variants, output_type);
+
         let try_from = quote!{
             impl#impl_generics ::std::convert::TryFrom<#input_type#ty_generics> for
                 (#(#original_types),*) #where_clause {
-                type Error = ();
+                type Error = &'static str;
 
                 #[allow(unused_variables)]
                 #[inline]
                 fn try_from(value: #input_type#ty_generics) -> Result<Self, Self::Error> {
                     match value {
                         #(#matchers)|* => Ok(#vars),
-                        _ => Err(()),
+                        _ => Err(#message),
                     }
                 }
             }

```

I'm testing something like this, which works, but I kind of hope I'm missing something to make the formatting code less horrible.

tests/try_into.rs
```diff
@@ -0,0 +1,79 @@
...
+    Unsigned(u32),
+    NamedUnsigned { x: u32 },
+    Unit,
+}
+
+#[test]
+fn test_try_into() {
+    let i = MixedInts::SmallInt(42);
+    assert_eq!(Ok(42i32), i.try_into());
+    assert_eq!(TryInto::<i64>::try_into(i), Err("Only NamedBigInt can be converted to i64"));
```
JelteF: I think these errors can probably be done using `i64::try_from()`. Like you did in the example. If so can you use that and also remove the `use std::convert::TryInto;` line.

src/lib.rs
```diff
@@ -107,6 +108,7 @@
 //! 10. [`#[derive(AddAssign)]`](add_assign.html)
 //! 11. [`#[derive(MulAssign)]`](mul_assign.html)
 //! 12. [`#[derive(Constructor)]`](constructor.html)
+//! 13. [`#[derive(TryInto)]`](try_info.html)
```
JelteF: should be `try_into.html`

JelteF: and can you place this Below `FromStr` in this list? I like to keep the same ordering of traits throughout the docs.

src/lib.rs
```diff
@@ -167,6 +169,7 @@
 //! [`RemAssign`]: https://doc.rust-lang.org/std/ops/trait.RemAssign.html
 //! [`ShrAssign`]: https://doc.rust-lang.org/std/ops/trait.ShrAssign.html
 //! [`ShlAssign`]: https://doc.rust-lang.org/std/ops/trait.ShlAssign.html
+//! [`TryInto`]: https://doc.rust-lang.org/core/convert/trait.TryInto.html
```
JelteF: Ordering here as well (though not actually visible to users).

JelteF:
Looks really good now! I added some final comments, if you address those it can be merged.

doc/try_into.md
```diff
@@ -0,0 +1,155 @@
+% What #[derive(TryInto)] generates
+
+This derive allows you to convert enum variants into their corresponding
+variant types.
+One thing to note is that this derive doesn't actually generate an
+implementation for the `TryInto` trait.
```
JelteF: Can you note somewhere here that this derive only works on the `nightly` compiler?

src/lib.rs
```diff
@@ -61,6 +61,7 @@
 //! 1. [`From`]
 //! 2. [`Into`]
 //! 3. [`FromStr`]
+//! 4. [`TryInto`]
```
JelteF: Can you put `(nightly only)` after the link?

doc/try_into.md
```diff
@@ -0,0 +1,155 @@
...
+Instead it derives `TryFrom` for each variant in the enum and thus has an
+indirect implementation of `TryInto` as recommended by the
+[docs](https://doc.rust-lang.org/core/convert/trait.TryInto.html).
+
+# Example usage
+
+```rust
+#![feature(try_from)]
+# #[macro_use] extern crate derive_more;
+use std::convert::TryFrom;
```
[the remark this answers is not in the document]
glandium: It's not in the prelude (yet?).

doc/try_into.md
```diff
@@ -0,0 +1,155 @@
...
+#[derive(TryInto)]
+enum MixedData {
+    Int(u32),
+    String(String),
+}
+
+fn main() {
+    let s = MixedData::String("foo".into());
+    assert_eq!("FOO", String::try_from(s).unwrap().to_uppercase());
```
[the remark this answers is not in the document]
glandium: Unfortunately not.

JelteF:
Thanks for all the work on this

JelteF:
I expanded the example a tiny bit btw

JelteF:
Just released v0.10.0 with these changes as well as support for deriving `Deref` and `DerefMut`.

PR #47, opened 2018-04-01
Update dependencies

hcpl:


JelteF:
Thanks. Nice that almost nothing broke this time. I've spent hours updating to the previous version of `syn`.

hcpl:
Heh. You made it sound like I was cheating with this one :laughing:.

Joking aside, I am glad too that `syn` and `quote` API breakages were mostly because of `TokenStream` and other types from `proc-macro2` being public dependencies that has actually changed a lot.

PR #55, opened 2018-09-29
Upgrade to syn 0.15 and extend Display-like derivations to support custom formats

Laaas:
Allows stuff like the following:
```rust
#[derive(Display)]
enum E {
    #[display(fmt = "A({:o})", "_0")]
    A(u32),
    #[display(fmt = "B {i: {:b})", "i + 29")]
    B {i: i32},
}
```

The only issue is that you have to wrap arguments in strings, even though they are expressions. This is a limitation with the rust syntax, but I find this better than what `failure` does (only allows identifiers) since it's tedious to display paths with `failure`.

There's 2 semi-hacks in my PR:
- Errors aren't always reported with proper spans, since some times rustc loses the information somehow (see [this](https://github.com/dtolnay/syn/issues/508))
- The output of each derivation function is extended to allow `Results`, by utilizing traits. Perhaps this should not be used, and each derivation function should always return a `Result` instead?

### DIFF
--- a/doc/display.md
+++ b/doc/display.md
@@ -41,10 +41,6 @@ extern crate derive_more;
 
 use std::path::PathBuf;
 
-// Here just to make sure that this doesn't conflict with
-// the derives in some way
-use std::fmt::Binary;
-
 #[derive(Display)]
 struct MyInt(i32);
 
--- a/doc/display.md
+++ b/doc/display.md
@@ -36,8 +36,7 @@ E.g. `Octal` -> `octal`, `Pointer` -> `pointer`, `UpperHex` -> `upper_hex`.
 # Example usage
 
 ```rust
-#[macro_use]
-extern crate derive_more;
+# #[macro_use] extern crate derive_more;
 
 use std::path::PathBuf;
 
@@ -90,8 +89,7 @@ struct Unit;
 #[derive(Display)]
 struct UnitStruct {}
 
-#[test]
-fn check_display() {
+fn main() {
     assert_eq!(MyInt(-2).to_string(), "-2");
     assert_eq!(Point2D { x: 3, y: 4 }.to_string(), "(3, 4)");
     assert_eq!(E::Uint(2).to_string(), "2");
--- a/doc/display.md
+++ b/doc/display.md
@@ -1,5 +1,7 @@
 % What #[derive(Display)] generates
 
+**NB: These derives are fully backward-compatible with the ones from the display_derive crate.**
+
 Deriving `Display` will generate a `Display` implementation, with a `fmt`
 method that matches `self` and each of its variants. In the case of a struct or union,
 only a single variant is available, and it is thus equivalent to a simple `let` statement.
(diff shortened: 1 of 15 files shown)


Laaas:
Well it seems like my code doesn't compile on stable rust because `error[E0658]: non-string literals in attributes, or string literals in top-level positions, are experimental (see issue #34981)`, but that issue is marked as resolved. Perhaps it will be available in 1.30?

Laaas:
Though it fails, actually supporting this as of yet unstable syntax would not be a breaking change, so you could still merge it.

JelteF:
Thanks for the work on this! I'll try to review this in the near future. 

Regarding your two last comments about unstable feature usage. Could you create a test file that tests the unstable features as well? You can use the same thing as here to disable the tests for stable rust: https://github.com/JelteF/derive_more/blob/master/tests/try_into.rs#L1

Laaas:
It is very possible, but I was thinking it may be confusing

Laaas:
Unfortunately, I had to remove all use of `quote_spanned!` since it didn't always work, and when it didn't work, it caused the same problem as with compile-time errors: it would not report what derive failed at all. I didn't find any way to validate that the returned tokenstream has valid spans (I think you can go through each token tree and debug print the spans, but that would probably be quite slow..).

Laaas:
Well I fixed the issue with spans by providing a wrapper over syn's Spanned trait, which checks the validity of the span, and in the case that it's invalid, just returns the default call_site span.

Laaas:
Also going to update the documentation now.

Laaas:
I pushed the work-around for spans into syn BTW, so they're not needed here anymore.

Laaas:
Hello, can you review this?

src/display.rs
```diff
@@ -1,55 +1,213 @@
...
+            })?;
+
+        Ok(quote_spanned!(meta.span()=> write!(_derive_more_Display_formatter, #fmt, #args)))
+    }
+    fn infer_fmt(&self, span: Span, fields: &Fields) -> Result<TokenStream> {
+        let fields = match fields {
+            Fields::Unit => {
+                return Err(Error::new(
+                    span,
+                    "Can not automatically infer format from unit type",
```
JelteF: Can you instead return the name of the Field here. I got a PR for that functionality in #56 and I think it's indeed quite reasonable, especially now that you can override it it with the `fmt="something"` that this PR adds.

src/display.rs
```diff
@@ -1,55 +1,213 @@
...
+                ))
+            }
+            Fields::Named(fields) => &fields.named,
+            Fields::Unnamed(fields) => &fields.unnamed,
+        };
+        if fields.len() != 1 {
```
JelteF: can you return the name of the field when it's a struct with 0 fields as well?

JelteF:
Also, please don't rebase/squash new changes with the previous commits. That makes it much easier for me to review your new changes on github.

doc/display.md
```diff
@@ -1,78 +1,69 @@
...
-# Enums
+#[derive(Display)]
+#[display(fmt = "Hello there!")]
+union U {
+    i: u32,
+}
 
-Deriving `Display` is not supported for enums.
+#[derive(Octal)]
+#[octal(fmt = "8")]
```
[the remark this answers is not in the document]
Laaas: This was a joke, but sure.

Laaas:
Thanks! Hadn't checked my e-mail lately, so hadn't seen this sadly. Will attempt to finish this ASAP.

Laaas:
I think I've fixed it all now. Please check again.

Laaas:
One of the tests are erring due to the `extern crate` in the doc/display.md file, but I can not reproduce this locally with `cargo external-doc`. It generates them all fine for me.

JelteF:
The changes look good!

Could you add this code that would not compile as a test:
```rust
trait Trait {
	fn display(&self) -> impl Display;
}

\#[derive(Display)]
\#[display(fmt = "{}", "t.display()")]
struct S<T: Trait> {
	t: T,
}
```

Regarding the doc tests that fail it's because I use a fork of cargo-external-doc that actually fails when doc tests don't pass. See this PR for details: https://github.com/Geal/cargo-external-doc/pull/4
To reproduce the failure locally you should do:
```
cargo install --git https://github.com/JelteF/cargo-external-doc --force
cargo build
cargo external-doc
```
(sometimes it starts failing for weird reasons, in that case try `cargo clean && cargo build` to refresh the build cache)

doc/display.md
```diff
@@ -1,78 +1,113 @@
...
 
+```rust
+#[macro_use]
+extern crate derive_more;
 
-# Regular structs
+use std::path::PathBuf;
 
+// Here just to make sure that this doesn't conflict with
+// the derives in some way
```
JelteF: can you remove this from the example it adds some noise

doc/display.md
```diff
@@ -1,78 +1,113 @@
...
+struct UH;
+
+#[derive(Display)]
+struct Unit;
+
+#[derive(Display)]
+struct UnitStruct {}
+
+#[test]
+fn check_display() {
```
JelteF: you probably have to change this to main to fix the doc tests

JelteF:
If you do these final things I will merge the PR btw

Laaas:
I think it's done now.

JelteF:
:+1: Thank you for your work on this. I'll create release soon-ish

JelteF:
Just released 0.13.0 with this feature.

PR #56, opened 2018-10-09
Impl display enum

matthiasbeyer:
Hi,

my attempt to provide a implementation to derive `Display` for enums, shamelessly ripped off of https://github.com/ihrwein/enum-display-derive (Apache2, MIT - so no problem I guess).

Does not work, though, the test fails with:

```
error[E0405]: cannot find trait `Display` in this scope
  --> tests/display.rs:14:10
   |
14 | #[derive(Display)]
   |          ^^^^^^^ not found in this scope
help: possible candidate is found in another module, you can import it into scope

```

Maybe you have an idea?

### DIFF
--- a/src/display.rs
+++ b/src/display.rs
@@ -9,22 +9,36 @@ pub fn expand(input: &DeriveInput, trait_name: &str) -> TokenStream {
     let generics = add_extra_ty_param_bound(&input.generics, trait_path);
     let (impl_generics, ty_generics, where_clause) = generics.split_for_impl();
     let input_type = &input.ident;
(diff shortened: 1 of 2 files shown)


JelteF:
Hi thanks for the PR, but I'm gonna close this in favor of #55, since it adds much more functionality. I have asked the creator of that PR to also add support for your use case.

matthiasbeyer:
You're awesome, thanks!

JelteF:
Just released 0.13.0 with the new Display derive, including a change that supports this feature.

Issue #58, opened 2018-11-08
Add no-std support


gbip:
Would it be possible to add `no-std` support for this crate ?

I am willing to implement it if it is not too complicated... :smile:

JelteF:
Afaik it shouldn't be very complicated. I think all that's needed is
outputting implementations of core versions of all the traits instead of
the std ones when the nostd feature is enabled. Feel free to make a PR with
the necessary changes.

gbip:
It seems that your are using a `HashMap` in [from.rs](https://github.com/JelteF/derive_more/blob/master/src/from.rs#L75) and there is no equivalent in `[no-std]`. Is this a blocking problem ?

JelteF:
I would assume the no-std would apply to the generated code only, not the code I use to generate it. AFAIK the Hashmap is only used to generate the code.

gbip:
Do you agree that what I should do is replace all the reference to std in the generated implementation (such as [here](https://github.com/JelteF/derive_more/blob/master/src/add_assign_like.rs#L28)) to references to either `core` or `std` based on the presence of the `no_std` feature ?

If I'm right then I thought about two solutions :
1.Wrap the call to `quote!` in a [cfg_if](https://docs.rs/cfg-if/0.1.6/cfg_if/) block everywhere but this will lead to duplicated code...
2. Create a macro that would wrap `quote!` and either spit out a `std` or `no_std` implementation.

Is the last solution a good way to solve this problem ?

JelteF:
I think that should be enough to make it work with no_std indeed.
Regarding the way to solve it, I think it's best to make the std/core part of the quote a variable that you define outside the full quote and then reference it inside the full quote (possibly in the root of the package). So something like this:

```
let import_root = if std {
     quote!(::std)
} else {
    quote!(::core)
} 
    quote!(
        impl#impl_generics #import_root::ops::#trait_ident for #input_type#ty_generics #where_clause {
            #[inline]
            fn #method_ident(&mut self, rhs: #input_type#ty_generics) {
                #(#exprs;
                  )*
            }
        }
)
```

gbip:
Do you have any idea about writing integration tests for this feature ?
Ideally we would run the test suite without the `no_std` feature and then with it but it does not seems possible at the moment...

JelteF:
The failure crate does it apparently:
https://github.com/rust-lang-nursery/failure/issues/30

PR #61, opened 2018-11-21
Added `no std` support

gbip:
I followed @JelteF guidance to implement `no_std` support for this crate (#58).
There might be a few thing that we have to do before merging :
* figure a way to test the `no_std` feature in local (currently it is tested on in CI)
* would mentioning `no_std` support in the crate description be ok for you ?


gbip:
Hey @JelteF could you take a look at this ? :smile:
I have a crate that depends on this to be published on crates.io.

JelteF:
Thanks for the PR. Could you change it such that the import_root is a global in the utils file? That way it doesn't need to infect the arguments of so many functions in the library. If a global doesn't work for whatever reason a function that returns the import root is also fine. But I think a global should be possible, probably lazy_static is needed though.

JelteF:
@gbip What do you think about the change I suggested?

gbip:
I am willing to apply your changes but I just don't have the time to do this right now... :/

JelteF:
No problem. Mainly wanted to make sure you saw my response.

PR #73, opened 2019-03-13
Fix unhygenic use of `Result` in `TryInto` macro

mooli:
Code such as this (untested) breaks with `wrong number of type arguments: expected 1, found 2` due to the `TryInto` macro using a bare `Result`:
```
pub type Result<T> = core::result.Result<T, Error>;
#[derive(TryInto)]
pub enum Error { ... }
```
This PR fixes that problem by fully-qualifying the name.


JelteF:
Thanks for the PR!

PR #78, opened 2019-05-30
Provide type parameter bounds for Display

### LINKED ISSUE #72: Better support for generics in Display
## Current behavior

```rust
use derive_more::Display;

#[derive(Display)]
enum MyError<E> {
    #[display(fmt = "Custom error: {}", _0)]
    Custom(E),
    #[display(fmt = "{}", _0)]
    Concrete(DbError),
}

#[derive(Display)]
#[display(fmt = "database error")]
struct DbError;
```

```
   |
48 |     #[display(fmt = "Custom error: {}", _0)]
   |                                         ^^ `E` cannot be formatted with the default formatter
   |
   = help: the trait `std::fmt::Display` is not implemented for `E`
   = note: in format strings you may be able to use `{:?}` (or {:#?} for pretty-print) instead
   = help: consider adding a `where E: std::fmt::Display` bound
   = note: required because of the requirements on the impl of `std::fmt::Display` for `&E`
   = note: required by `std::fmt::Display::fmt`
```

With `enum MyError<E: std::fmt::Display> {` works fine.

## Desired behavior

As std `#[derive(Debug)]` does: imply `std::fmt::Display` bound on `impl` not the type.

```rust
#[derive(Debug)]
enum MyError<E> {
    Custom(E),
    Concrete(DbError),
}

#[derive(Debug)]
struct DbError;
```

[Compiles OK](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=19260ad4a9cae15099966d6b26331d78).

## Motivation

Traits bounds on the type itself pollute all the places where the type is used with these bounds (requires to specify them anywhere), which makes things quite unorthogonal. Its better to have trait bounds on `impl`s, so that user may decide whenever he wants these behavior or not in generic code by himself, rather than pushing him to apply this traid bound everywhere on type parameters.

tyranron:
Fixes #72

## Background

See #72 for the detailed issue description and motivation.

```rust
#[derive(Display)]
#[display(fmt = "{} {} {{0:p}} {0:o} {1:#x} - {0:>4?} {1:^4X?}", "_0", "_1")]
struct MultiTraitUnnamedGenericStruct<A, B>(A, B);
```
expands to
```rust
impl<A, B> ::std::fmt::Display for MultiTraitUnnamedGenericStruct<A, B>
where
    A: ::std::fmt::Display + ::std::fmt::Debug + ::std::fmt::Octal,
    B: ::std::fmt::Display + ::std::fmt::Debug + ::std::fmt::LowerHex,
{
    // ...
}
```

## Solution

- Apply the necessary trait bounds to type parameters __only if they are used__ in formatting expression.
- Detect the correct trait bound (`Display`, `Debug`, `Octal`, etc.) basing on [formatting placeholder format](https://doc.rust-lang.org/stable/std/fmt/index.html#syntax).
- Apply the necessary trait bounds for automatically inferred implementations (when no `#[display]` attribute is provided).

## Limitations

- `union`s are not supported (are generic `union`s really used?).
- `enum` with `#[display]` attribute on itself (not on its variant) are untouched, as current deriving implementation doesn't allow using fields for this case.
- Expressions using fields (such as `"_0.as_str()"` or `"2 * field + field"`) are not supported. It turns out that without a compile-time reflection it's impossible to detect the result type of expression correctly in all cases, so we cannot know reliably whether we should apply trait bound to the type param field used in an expression or not. To avoid false positives I decided to leave this case "as is", so it works the same way the current implementation does.

## Checklist

- [x] Code formatted
- [x] Clippy is happy
- [x] Tests are added
- Documentation is not updated (nothing is changed in usage API)
- Changelog entry is not added (seems that you manage these by yourself while releasing)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,9 +25,11 @@ name = "derive_more"
 proc-macro = true
 
 [dependencies]
+lazy_static = "1.3"
+proc-macro2 = "0.4"
 quote = "0.6"
-syn = { version = "0.15", features = ['extra-traits'] }
-proc-macro2 = "0.4.19"
+regex = "1"
+syn = { version = "0.15", features = ["extra-traits"] }
 
 [build-dependencies]
 rustc_version = "0.2"
--- a/src/display.rs
+++ b/src/display.rs
@@ -226,14 +226,14 @@ impl<'a, 'b> State<'a, 'b> {
                             let name = &self.input.ident;
                             let v_name = &v.ident;
                             let fmt: TokenStream;
-                            let mut bounds = HashMap::new();
+                            let bounds: HashMap<_, _>;
 
                             if let Some(meta) = self.find_meta(&v.attrs)? {
                                 fmt = self.get_meta_fmt(&meta)?;
                                 bounds = self.get_used_type_params_bounds(&v.fields, &meta);
                             } else {
                                 fmt = self.infer_fmt(&v.fields, v_name)?;
-                                // TODO: bounds?
+                                bounds = self.infer_type_params_bounds(&v.fields);
                             };
                             all_bounds = bounds.into_iter()
                                 .fold(all_bounds, |mut bounds, (ty, trait_names)| {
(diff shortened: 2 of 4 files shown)


tyranron:
ping @JelteF

JelteF:
Thanks a lot for your work on this. This looks like a great addition. I'll make a new release with this right away.

PR #79, opened 2019-06-10
[WIP] Implement proof-of-concept of an enum "affix"

ErichDonGubler:
It'd be really nice to be able to derive `Display` and use an affix (prefix and/or suffix) that comes before each inner implementation of `Display`. This allows one to write a `derive`d implementation of `Display` like so:

```rust
    // Expects a single Display item corresponding to where an inner 
    // variant's Display will be written (`{}`), otherwise errors
    #[derive(Debug, Display)]
    #[display(affix = "unable to read from /proc/stat: {}")]
    pub(crate) enum SnapshotReadError {
        Io(IoError),
        #[display(fmt = "expected CPU stat line for {}, got {:#?}", expected_cpu, line)]
        UnableToParseLine {
            expected_cpu: CpuComponent,
            line: String,
        }
    }

    #[derive(Debug, Display)]
    pub(crate) enum CpuComponent {
        #[display(fmt = "entire CPU")]
        EntireCpu,
        #[display(fmt = "CPU core {}", _0)]
        Core(usize),
    }
```

This can be particularly handy for error handling, but I'm sure there are other places this could be applied as well. :)

---

Things to do before this should be merged:

- [ ] Design feedback: Mostly, is this interesting?
- [ ] This commit probably shouldn't make things as hard to follow as it does -- how could this improve?
- [ ] I specifically chose a terrible but concise word for this which I fully expect to be bikeshedded. Alternatives might include:
    * `outer_fmt` (I like this one the best.)
    * `prefix` (but it's not just a prefix!)
    * `surround`

### DIFF
--- a/src/constructor.rs
+++ b/src/constructor.rs
@@ -21,7 +21,7 @@ pub fn expand(input: &DeriveInput, _: &str) -> TokenStream {
         _ => panic!("Only structs can derive a constructor"),
     };
     let original_types = &get_field_types(&fields);
-    quote!{
+    quote! {
         #[allow(missing_docs)]
         impl#impl_generics #input_type#ty_generics #where_clause {
             #[inline]
--- a/src/deref.rs
+++ b/src/deref.rs
@@ -28,7 +28,7 @@ pub fn expand(input: &DeriveInput, trait_name: &str) -> TokenStream {
     let (impl_generics, ty_generics, where_clause) = generics.split_for_impl();
     // let generics = add_extra_ty_param_bound(&input.generics, trait_path);
     let casted_trait = &quote!(<#field_type as #trait_path>);
-    quote!{
+    quote! {
         impl#impl_generics #trait_path for #input_type#ty_generics #where_clause
         {
             type Target = #casted_trait::Target;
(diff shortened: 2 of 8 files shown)


JelteF:
1. I think this is a nice addition
2. It's not too crazy, sadly code generation code is in generally quite hard to follow.
3. If possible it would be very nice if we could simply use `fmt`. This would require a bit of changes to your code though, but I think it might actually make it simpler. If this is not possible I agree that `outer_fmt` is the best choice.

JelteF:
I changed the PR such that you can use `fmt` this way. It's on master now.

JelteF:

ErichDonGubler:
@JelteF: After using this in production for some time, I want to submit that an `affix` key (as originally proposed in the OP I wrote) was probably the right design in the first place. A year after this gets merged, I came back to use a top-level `enum` format, something of the form:

```rust
use derive_more::Display;

#[derive(Debug, Display)]
#[display(fmt = "{:?}", self)]
enum Enum {
    Variant1,
    // ...
}
```

...but was confused by the diagnostic given (which I've [filed a PR for][confusing-diagnostic-pr]) about a single `fmt` needing a single argument. It was only after examining source that I remembered that a top-level `#[display(fmt = ..., ...)]` on an `enum` uses an "affix mode". It's not very clear that suddenly expectations have changed, and it also makes the (somewhat intuitive, I think) case I was trying to implement above _impossible_ without just manually `impl`ing the desired `std::fmt` trait again. Furthermore, I think it makes much more sense to prefer the above behavior for an outer `enum` `fmt` string, since that's much closer to what is already possible for 

I'll filed a [new issue][new-issue] proposing a breaking change for this use case with the above information, so let's continue discussion there.

[confusing-diagnostic-pr]: https://github.com/JelteF/derive_more/pull/140
[new-issue]: https://github.com/JelteF/derive_more/issues/142

Issue #48, opened 2018-04-04
Difference in derive(Deref) from derive_deref


spease:
Looks like this crate derives deref based on the enclosed item's implementation of deref, whereas derive_deref sets up deref to deref to the enclosed item.

EG where
```
B::deref() -> &C

#[derive(Deref)]
struct A(B);
```

derive_deref
```
A::deref -> &B
```

derive_more
```
A::deref -> &C
```

For the code I've worked with, the former case is usually the more commonly desired one since I've often had to use wrapper structs for serialization that should otherwise be transparent.

I'm not sure if that's the absolute most common desired implementation though.

JelteF:
I understand that your usecase makes sense, but I don't really think it fits well in this library. For derive `Deref` usecase I actually like the more explicit derive `Into` that this lib provides better for this behaviour. However, I don't think you could also use it as an equivalent for `DerefMut`.

I hadn't seen derive_deref before though btw. It does seem to work both for my current implementation, dereferencing twice when the inner type implements `Deref` as well, but having the other behaviour otherwise. I had seen [shrinkwraprs]( https://crates.io/crates/shrinkwraprs) before though, which does sort of the same, but derives some more traits and can also use some attributes to choose field to deref. 
Because that one does `derive(Shrinkwrap)` you can also use that sometimes and still use my implementation for the cases where you want to deref twice.

I will leave this issue open for now, because the docs for `Deref` and `DerefMut` should contain this information as well.

spease:
So the recommended solution here is to use shrinkwrap for those situations where the struct is a transparent wrapper sort of object?

General cases where this comes up is for custom serialization, and forcing type safety although the data storage is still a string, u64, etc.

JelteF:
My advice would be:
1. use `Into` if you don't need mutable access
2. derive `Deref` and `DerefMut` if what you wrap is a pointer type, e.g `&`, `Rc` or `Box` 
3. Otherwise use shrinkwraprs

jsdw:
I think I just bumped into the same issue (using 0.12.0); I was surprised to find that

```
#[derive(Deref)]
struct Wrapper(u64);
```

leads to the error:

```
error[E0277]: the trait bound `u64: std::ops::Deref` is not satisfied
  --> src/main.rs:21:10
   |
21 | #[derive(Deref)]
   |          ^^^^^ the trait `std::ops::Deref` is not implemented for `u64`
```

[...]

PR #80, opened 2019-07-19
Issue #48: derive `Deref` and `DerefMut` directly on any newtype structs

magnet:
Derive `Deref` and `DerefMut` directly on any newtype structs's single field instead of requiring inner field to implement the traits.

This is a backwards-incompatible change, though many usages will require no call-site modification.

See issue #48 for discussion.

### DIFF
--- a/src/deref.rs
+++ b/src/deref.rs
@@ -27,14 +27,13 @@ pub fn expand(input: &DeriveInput, trait_name: &str) -> TokenStream {
     let generics = add_extra_ty_param_bound(&input.generics, trait_path);
     let (impl_generics, ty_generics, where_clause) = generics.split_for_impl();
     // let generics = add_extra_ty_param_bound(&input.generics, trait_path);
-    let casted_trait = &quote!(<#field_type as #trait_path>);
     quote!{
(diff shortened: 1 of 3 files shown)


JelteF:
I've thought about it and I don't really like that it's a breaking change like this. So I'm not gonna merge it like this. If you can make it non breaking in a nice way I might merge that, otherwise I think I want to wait for specialization to make both work with a single dereference operator.

magnet:
No problem.  I'm not sure how specialization will help here given what I understand of overlapping rules; as it is I don't think the breakage is big (though it would justify a version bump), because most people who use `Deref` do it using the dereference operator `.` rather than explicitly with `*` but it's your call.

JelteF:
Thinking about it a bit more I think it's a good addition. If you can change it to be a new derive I will merge it. My suggestions for the names are `DerefSelf` and `DerefMutSelf`. If you have better ones feel free as well.

JelteF:
Another suggestion for the name I came up with is `DerefToInner` and `DerefMutToInner`. I think they are clearer than `DerefSelf`.

JelteF:
I got this onto master

JelteF:

PR #82, opened 2019-08-07
Fixes #81

### LINKED ISSUE #81: Conditionally compile features
As this crate continues to expand, compilation time increases, but for a lot of cases, I simply don't want most of the macros available. (Although a release build of the current master only took 17.88 seconds on my CPU, and the derives aren't linked into the final executable anyway)

This issue suggests adding a feature for each derive macro.

SOF3:
## Changes
- Each module has its corresponding feature. Clustered derive macros like `add_like` macros share the same feature.
- `lazy_static` is only required when `feature(display)` is enabled
- Added a `#![allow(dead_code)]` tag for `src/util.rs` to prevent dead code warnings when features are disabled. This is not so elegant, but at least better than writing a `#[cfg(any(feature = "add_like", feature = "add_assign_like", ...))]` attribute for each function in `util`.

## Backward compatibility
All new features are enabled by default, so this should not affect existing usages

## Test
`cargo test --no-default-features` is failing for reasons I don't understand. Help needed.

## Docs
`cargo doc` is passing. I am not testing with feature flags since that's usually unimportant anyway.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,7 @@ matrix:
     - env: TARGET=x86_64-unknown-linux-gnu # this is the default job
       before_script:
         - rustup component add clippy
-        - go get go get github.com/pelletier/go-toml/cmd/tomljson
+        - go get github.com/pelletier/go-toml/cmd/tomljson
       script:
          - cargo clippy --all -- -D clippy::all
          - bash ci/test_all_features.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,9 +49,10 @@ matrix:
       rust: nightly
     - env: TARGET=x86_64-unknown-linux-musl DISABLE_TESTS=true
       rust: nightly
-    - env: TARGET=x86_64-apple-darwin
-      os: osx
-      rust: nightly
+      # Currentl failing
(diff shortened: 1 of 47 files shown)


JelteF:
I think this is a really nice addition! I have a couple of points:

1. Maybe I don't understand the changes correctly, but it seems like you're disabling tests when the correct feature is not enabled.
I think it makes more sense to do the same that is done for the `no_std` test and set required-features for each test like so in Cargo.toml:
```toml
[[test]]
name = "deref"
path = "tests/deref.rs"
required-features = ["deref"]
```

2. regex is only required for display, just like lazy_static
3. I think fixing number 1 will probably lead you to how to fix the test run for `cargo test --no-default-features`. It would be good to have this "test run" in CI by adding it to `ci/script.sh`. ("test run" in quotes because it will not run any tests)

JelteF:
1. I fixed for some tests in the last commit I pushed
2. Is not relevant anymore as dependencies on lazy_static and regex are removed
3. I fixed this, it were the doctests that couldn't compile.

JelteF:

SOF3:
I don't know why I decided to use snake_case instead of PascalCase for feature names...

PR #83, opened 2019-08-08
[WIP] Removes regex and lazy_static dependencies

kardeiz:
This PR replaces the regexes used to parse format placeholders with a custom parser, generated with [`rust-peg`](https://github.com/kevinmehall/rust-peg/tree/0.5). 

`rust-peg` provides a binary to convert a `.rustpeg` file to an `.rs` file. I'm checking in both `parsing.rustpeg` file, as well as `parsing.rs`, so as not to not introduce a dependency on `rust-peg`. When updating `parsing.rustpeg`, it will be necessary to run `rust-peg src/parsing.rustpeg > src/parsing.rs`.

This is a little bit clunky, but might be better than the alternatives: the regex dependency on the one hand, and a completely handrolled, complex parser for the format string on the other.

All tests and doc tests are passing.

Feel free to close if you plan on using `regex` elsewhere in the crate, or if you aren't worried about the size of the `regex` dependency. Thanks for `derive_more`&mdash;it's great!


JelteF:
I think this makes sense. Nice to not have dependencies if it can be reasonably be avoided. @tyranron could you look at this PR?

@kardeiz I haven't looked at this in detail yet, but one thing comes to mind already. Could you check if it's possible to add a development dependency on peg in some way. It would be nice if the parsing file has to be regenerated that we have a major version of peg pinned in Cargo.lock.

kardeiz:
@JelteF Yeah, I was thinking about that. My first thought was to have an `example` that would generate the `parsing.rs` file, but that would be too weird and confusing. (It would be nice if there were a generic `dev` phase in cargo not specifically `test`, `bench`, or `example`.)

A less weird thing I thought of was to have a feature flag that would require rust-peg that would conditionally regenerate `parsing.rs` in `build.rs` when enabled. One could then do `cargo build --features generate-parsing-rs` to generate the file. 

But it is probably best just to add `rust-peg` into `dev-dependencies` as you suggest, unless you or others have some other idea.

JelteF:
I think this is a really good solution:

>A less weird thing I thought of was to have a feature flag that would require rust-peg that would conditionally regenerate parsing.rs in build.rs when enabled. One could then do cargo build --features generate-parsing-rs to generate the file.

kardeiz:
@JelteF I've added the feature flag `generate-parsing-rs` as discussed above. The `parsing.rs` file can be generated with `cargo build --features generate-parsing-rs`.

JelteF:
Hmm, running this locally I get a lot of warnings like this:

```
warning: `...` range patterns are deprecated
   --> src/parsing.rs:278:32
    |
278 |                             '0'...'9' => Matched(__next, ()),
    |                                ^^^ help: use `..=` for an inclusive range
    |
    = note: #[warn(ellipsis_inclusive_range_patterns)] on by default
```

Can you make sure these warnings are `allow`-ed?

JelteF:
@tyranron Thank you for the second pair of eyes :)

PR #95, opened 2019-10-21
WIP: Improve derive-Display trait bounds inference (#93)

ffuugoo:
First half of #93.

Modified `Display` macro, so that for each field of the following types _or of reference to 
such types_ a where clause is generated bounding field's type by `Display` trait:

* `T`
* `T::AssociatedType` and `<T as Trait>::AssociatedType`
* `::path::SomeOtherType<T>` and `::path::SomeOtherType<&T>`

E.g., for a structure `Foo` defined like this:

```Rust
#[derive(Display)]
struct Foo<'a, T1, T2: Trait, T3> {
    field1: T1,
    field2: <T2 as Trait>::Type,
    field3: Bar<T3>,
    field4: &'a T1,
    field5: &'a <T2 as Trait>::Type,
    field6: &'a Bar<T3>,
}
```

Following where clauses would be generated (notice, that for reference types the 
_referred_ type is bound):

* `T1: Display`
* `<T2 as Trait>::Type: Display`
* `Bar<T3>: Display`

_As far as I understand,_ as we can't define `Display` for pointers, arrays, slices and tuples, adding bounds for such types (or any generic types parametrized by such types, e.g., `SomeType<[T; 10]>`) would be wrong.

Issue #93, opened 2019-10-14
Support custom bounds for Display-like traits


tyranron:
## Synopsis

Current `#[derive(Display)]` is smart enough with its default behavior. However, due to absence of compile-time reflection, it's impossible to handle all the cases really gracefully.

Consider the following:
```rust
#[derive(Display)]
enum Foo<A, B> {
    A(A),
    B(B),
}

#[derive(Display)]
enum Bar<A, B, C> {
    Foo(Foo<A, B>),
    C(C),
}
```

The `Bar` won't compile with an error:
```
`A` cannot be formatted with the default formatter`
```

This is because we're unable to infer `Display` requirements for `Foo<A, B>` in `Foo()` variant knowing only tokens.

Fortunately, such situations are tend to be resolved with custom `bound` attributes (see `#[serde(bound = "T: Serialize")]`, for example) in Rust ecosystem.

## Proposal

Let's extend `#[derive(Display)]` with `#[display(bound)]` attribute, which may be used in the following manner:

```rust
#[derive(Display)]
enum Foo<A, B> {
    A(A),
    B(B),
}

#[derive(Display)]
#[display(bound = "A, B")]
enum Bar<A, B, C> {
    Foo(Foo<A, B>),
    C(C),
}

// Expands to:
impl<A, B, C> fmt::Display for Bar<A, B, C>
where A: Display,
      B: Display,
      C: Display,
{ ... }
```
If bound is not specified (no semicolon), we apply the bound of the derived trait.

```rust
#[derive(Display)]
enum Foo<A, B> {
    A(A),
    #[display(fmt = "{0:?} + {0:o}", _0)]
    B(B),
}

#[derive(Display)]
#[display(bound = "A, B: Debug + Octal")]
enum Bar<A, B, C> {
    Foo(Foo<A, B>),
    C(C),
}

// Expands to:
impl<A, B, C> fmt::Display for Bar<A, B, C>
where A: Display,
      B: Debug + Octal,
      C: Display,
{ ... }
```

JelteF:
I believe I understand the problem, but I think a much simpler and more elegant solution would be to derive:
```rust
impl<A, B, C> fmt::Display for Bar<A, B, C>
where Foo<A, B>: Display,
      C: Display,
{ ... }
```

It could be that I'm misunderstanding though.

JelteF:
To confirm my solution works as well I just tried the following code and it compiles (without the where it doesn't):
```rust
#[derive(Display)]
[...]

PR #97, opened 2019-10-23
Implement display-bound attribute for derive-Display macro (#93)

ffuugoo:
Second half of #93. This PR is based on top of #95.

Implemented `[display(bound = "...")]` attribute for `Display` macro as proposed by @tyranron in #93.

### DIFF
--- a/doc/display.md
+++ b/doc/display.md
@@ -33,13 +33,12 @@ i.e. `_0`, `_1`, `_2`, etc.
 The syntax does not change, but the name of the attribute is the snake case version of the trait.
 E.g. `Octal` -> `octal`, `Pointer` -> `pointer`, `UpperHex` -> `upper_hex`.
 
-# Deriving `Display` for generic data types
+# Generic data types
 
 When deriving `Display` (or other formatting trait) for a generic struct/enum, all generic type 
 arguments used during formatting are bound by respective formatting trait.
 
-E.g., for a structure Foo defined like this:
-
+E.g., for a structure `Foo` defined like this:
 ```rust
 #[macro_use] extern crate derive_more;
 
@@ -57,22 +56,21 @@ struct Foo<'a, T1, T2: Trait, T3> {
 }
 ```
 
-Following where clauses would be generated:
-
+The following where clauses would be generated:
 * `T1: Display + Pointer`
 * `<T2 as Trait>::Type: Debug`
 * `Bar<T3>: Display`
 
-## Specifying additional trait bounds
+## Custom trait bounds
 
 Sometimes you may want to specify additional trait bounds on your generic type parameters, so that they
 could be used during formatting. This can be done with a `#[display(bound = "...")]` attribute.
 
-`#[display(bound = "...")]` accepts single string argument in a format similar to a format 
+`#[display(bound = "...")]` accepts a single string argument in a format similar to the format 
 used in angle bracket list: `T: MyTrait, U: Trait1 + Trait2`.
 
 Only type parameters defined on a struct allowed to appear in bound-string and they can only be bound
-by traits, i.e., no lifetime parameters or lifetime bounds allowed in bound-string.
+by traits, i.e. no lifetime parameters or lifetime bounds allowed in bound-string.
 
 ```rust
 #[macro_use] extern crate derive_more;
--- a/doc/display.md
+++ b/doc/display.md
@@ -41,12 +41,18 @@ arguments used during formatting are bound by respective formatting trait.
 E.g., for a structure Foo defined like this:
 
 ```rust
+#[macro_use] extern crate derive_more;
+
+trait Trait {
+    type Type;
+}
+
 #[derive(Display)]
-#[display(fmt = "{} {:?} {} {:p}", a, b, c, d)]
+#[display(fmt = "{} {} {:?} {:p}", a, b, c, d)]
 struct Foo<'a, T1, T2: Trait, T3> {
     a: T1,
     b: <T2 as Trait>::Type,
-    c: Bar<T3>,
+    c: Vec<T3>,
     d: &'a T1,
 }
 ```
@@ -62,16 +68,22 @@ Following where clauses would be generated:
 Sometimes you may want to specify additional trait bounds on your generic type parameters, so that they
 could be used during formatting. This can be done with a `#[display(bound = "...")]` attribute.
 
-`#[display(bound = "...")]` accepts single string argument in a format generally similar to a format 
+`#[display(bound = "...")]` accepts single string argument in a format similar to a format 
 used in angle bracket list: `T: MyTrait, U: Trait1 + Trait2`.
 
 Only type parameters defined on a struct allowed to appear in bound-string and they can only be bound
 by traits, i.e., no lifetime parameters or lifetime bounds allowed in bound-string.
 
 ```rust
+#[macro_use] extern crate derive_more;
+
+trait MyTrait {
+    fn my_function(&self) -> i32;
+}
+
 #[derive(Display)]
 #[display(bound = "T: MyTrait, U: ::std::fmt::Display")]
-#[display(fmt = "{} {}", "a.my_function()", "transform(b.to_string())")]
+#[display(fmt = "{} {}", "a.my_function()", "b.to_string()")]
 struct MyStruct<T, U> {
     a: T,
     b: U,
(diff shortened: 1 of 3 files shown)


doc/display.md
```diff
@@ -33,6 +33,54 @@ i.e. `_0`, `_1`, `_2`, etc.
...
+
+Specifying type argument without explicitly specifying trait bounds is a shortcut to bind by formatting 
+type.
+
+Only type parameters defined on a struct allowed to appear in bound-string and they can only be bound
+by traits, i.e., no lifetime parameters or lifetime bounds allowed in bound-string.
+
+```rust
+#[derive(Display)]
+#[display(bound = "T: MyTrait, U")]
```
JelteF: I think it's not obvious that `U` has to implement `Display` when reading this code. Can you remove the featuer that it automatically does this and require manually doing `U: Display`.

tests/display.rs
```diff
@@ -339,4 +339,50 @@ mod generic {
...
         }
     }
+
+    mod bound {
+        use super::*;
+
+        #[test]
+        fn bound_simple() {
+            #[derive(Display)]
+            #[display(bound = "T1, T2")]
```
JelteF: Aren't these bounds already automatically inferred in this case?

doc/display.md
ffuugoo: Done.

tests/display.rs
ffuugoo: They are. It was kinda my point, to ensure that explicitly specifying already inferred bounds won't break stuff, but I should have made it a separate test with better name. Fixed it in latest commit.

ffuugoo:
> I've squashed your other PR and merged in into master. Can you update this branch to use master as a base (`git cherry-pick` should do it).

Done.

ffuugoo:
@tyranron

src/display.rs
```diff
@@ -139,26 +201,107 @@ impl<'a, 'b> State<'a, 'b> {
...
+            .filter(|meta| {
+                match meta {
+                    Meta::List(MetaList {
+                        path: Path {
+                            segments: outer_meta_path,
+                            ..
+                        },
+                        nested: inner_meta,
+                        ..
+                    }) => {
```
tyranron: I guess this nested beast could be rewritten as we've discussed.

src/display.rs
```diff
@@ -1,35 +1,51 @@
...
+    NestedMeta,
+    Path,
+    PathArguments,
+    PathSegment,
+    Token,
+    TraitBound,
+    TraitBoundModifier,
+    Type,
+    TypeParamBound,
+    TypePath,
```
tyranron: Why import all this? Is using `syn::Type` or `syn::TypeParamBound` is too bad and requires `syn::` part to be removed?

This has sense if there is a really long paths, to shorten it for better readability. The `punctuated::Punctuated` is a good example. However, in much cases `syn::Something` is quite good enough.

src/display.rs
```diff
@@ -99,6 +114,50 @@ pub fn expand(input: &DeriveInput, trait_name: &str) -> Result<TokenStream> {
     })
 }
 
+fn trait_name_to_attribute_name(trait_name: &str) -> &'static str {
```
tyranron: Consider to use smart well-known abbreviations, like `attr` instead `attribute`.

tests/display.rs
```diff
@@ -339,4 +339,60 @@ mod generic {
             assert_eq!(s.to_string(), "10");
         }
     }
+
+    mod bound {
+        use super::*;
+
+        #[test]
+        fn bound_simple() {
```
tyranron: No need to name functions with `bound_` prefix if they are placed inside `bound` module already.

src/display.rs
ffuugoo: Just followed existing style here. I'm for explicitly prefixing `syn::` personally.

tyranron: @JelteF wouldn'y you mind on some cosmetic refactoring for `src/display.rs` file?

JelteF: Sounds good for the changes made here, but if it touches code in another part of the file let's do that in another PR.

src/display.rs
```diff
@@ -139,37 +180,105 @@ impl<'a, 'b> State<'a, 'b> {
...
+        if bounds.is_empty() {
+            return Err(Error::new(span, "No bounds specified"));
+        }
+
+        bounds.into_iter().try_fold(HashMap::new(), |mut accumulator, generic_param| {
+            match generic_param {
+                syn::GenericParam::Type(type_param) => {
+                    if self.type_params.contains(&type_param.ident) {
+                        if !type_param.attrs.is_empty() {
+                            Err(Error::new(span, "Attributes aren't allowed"))
```
JelteF: Can you refactor this deeply nested code to use early returns instead? See this for a good example:
https://github.com/JelteF/derive_more/blob/master/src/utils.rs#L613-L633

src/display.rs
tyranron: @ffuugoo ☝️

src/display.rs
ffuugoo: Done.

src/display.rs
```diff
@@ -359,32 +469,71 @@ impl<'a, 'b> State<'a, 'b> {
...
                     quote_spanned!(self.input.span()=> _ => write!(_derive_more_Display_formatter, "{}", #fmt),),
                     HashMap::new(),
                 ))
             }
-        }
+        };
+
+        let (arms, mut bounds) = result?;
+
+        match self.find_meta(&self.input.attrs, "bound")? {
```
JelteF: Can you change this to early returns too?

ffuugoo: Also done.

ffuugoo:
That's what you get when you use nightly by default. Should be fixed now.

doc/display.md
```diff
@@ -33,10 +33,65 @@ i.e. `_0`, `_1`, `_2`, etc.
...
+```rust
+#[macro_use] extern crate derive_more;
+
+trait MyTrait {
+    fn my_function(&self) -> i32;
+}
+
+#[derive(Display)]
+#[display(bound = "T: MyTrait, U: ::std::fmt::Display")]
+#[display(fmt = "{} {}", "a.my_function()", "b.to_string()")]
```
JelteF: Let's add a function call around `b.to_string`, because now this bound is not necessary.

ffuugoo: It's actually required. As we specify "expression" (i.e., string literal) as a fmt argument, all bets are off. No default trait bound is inferred for `U`. But I'll change this example to something more obvious.

JelteF: It's a shame that that's not working. Is it possible to keep the automatic detection of bounds and merge them with the supplied bounds? That way you wouldn't have to supply the Display bound in this case.

JelteF: Oh wait I think I misunderstood your message. You meant that this works:
```rust
#[fmt="{}", b]
```
but this doesn't?
```rust
#[fmt="{}", "b"]
```

ffuugoo: Exactly. Was in the process of writing a better clarification. (:

ffuugoo: I can change example to something like `b.to_string().len()` (as it's a more useful example) and add a clarification to documentation, that for generic parameters in cases like `#[fmt="{}", "b"]` meaningful trait bound inference is impossible.

ffuugoo: (I assume we're speaking about the most trivial case of just `"<field-name>"`, cause, as written in documentation, string literals in `fmt` arguments are treated as arbitrary-rust-expression-that-is-just-pasted-into-generated-implementation-as-is and we can't meaningfully infer any bounds here.)

As far as I can say, it would not be overly complex, but not absolutely trivial either. And I think such change is out of scope for this PR: implementing such bound-detection would take place in `fmt` parsing code and I've never touched that code here.

I'd also argue if implementing this feature would be reasonable at all. We already have an all-around nicer way to achieve the same result: just write `<field-name>`. It's less typing and simpler macro implementation. There are still limitations to what you can (and what you should) achieve with derive-proc-macro. (:

ffuugoo: @JelteF I've updated the docs.

tyranron: @ffuugoo whenever we parse code of expression or not it doesn't matter, because we don't have type information of this expression. To support such cases proc_macros are not enough and requires support of compile-time reflection in Rust.  
However, supporting tivial case like `"<field-name>"` seems feasible.

tyranron: @JelteF @ffuugoo but I still don't think this is a good choice, because now we loose the difference between `a` field and `"a"` literal.

JelteF: Let's keep it like this then

JelteF:
I think it's good now. It's still a draft PR. Is that on purpose? If no, I will merge this PR.

PR #108, opened 2019-11-16
Refactor derive(Display) to properly propagate the alternate flag

### LINKED ISSUE #107: derive(Display) does not propagate the alternate flag
In some of my code, I've been using the alternate flag `{:#}` to modify some formatting. I've been relying on the fact that deriving Display on an enum would propagate this flag downwards. (same for Binary, Octal, LowerHex, UpperHex, LowerExp, UpperExp, Pointer)

However, sometime betwee versions 0.15.0 and 0.99.1, this seems to have changed. (A quick bisect has led me to the commit d8de4b80.)

Minimal reproduction:
```rust
use derive_more::{Display, Binary, Octal};

#[derive(Display, Binary, Octal)]
struct MyInt(u32);

#[test]
fn check_display() {
    assert_eq!(format!("{:b}", MyInt(9)), "1001");
    assert_eq!(format!("{:#b}", MyInt(9)), "0b1001");
}
```
Using the current master, the second line currently fails because `MyInt(9)` gets formatted as "1001" instead.

Has this change been made on purpose? If not, what would be the best way to solve this issue? I don't have a lot of experience with proc-macros, but I'm sure with a few pointers, I could put together a PR.

troiganto:
Closes #107 

After having a look into the expanded code and wrapping my head around, I finally found the issue. the `Display` derive uses a helper struct for the case of an enum with nested format-strings like the following (taken from the tests):

```rust
#[derive(Display)]
#[display(fmt = "Here's a prefix for {} and a suffix")]
enum Affix {
    A(u32),
    #[display(fmt = "{} -- {}", wat, stuff)]
    B {
        wat: String,
        stuff: bool,
    },
}
```

The problem was that if there was no outer format string, the default `"{}"` would be used, which removes all formatting settings from properly propagating.

A quick workaround was to avoid the treacherous line `write!(f, "{}", helper_object)` and instead call `Display::fmt(&helper_object, f)` directly, so that's what I did in 2c74515. After I slowly became more familiar with the code, I actually managed to completely avoid the helper object when it isn't necessary (c9c1ada). Ultimately, I even managed to only generate the type definition for this helper object when it is really necessary (faf2e84).

Once that was done, I couldn't help myself and cleaned up a few places that I found confusing, e.g. 2b0ff54. :slightly_smiling_face: 

Thank you for providing and maintaining this library and thank you for the thorough test suite: it really sped up the process and helped me make sure that my changes didn't break anything!

### DIFF
--- a/src/deref.rs
+++ b/src/deref.rs
@@ -3,7 +3,7 @@ use proc_macro2::TokenStream;
 use quote::quote;
 use syn::{parse::Result, DeriveInput};
 
-/// Provides the hook to expand `#[derive(Index)]` into an implementation of `From`
+/// Provides the hook to expand `#[derive(Deref)]` into an implementation of `Deref`
 pub fn expand(input: &DeriveInput, trait_name: &'static str) -> Result<TokenStream> {
     let state = State::with_field_ignore_and_forward(
         input,
--- a/src/deref_mut.rs
+++ b/src/deref_mut.rs
@@ -3,7 +3,7 @@ use proc_macro2::TokenStream;
 use quote::quote;
 use syn::{parse::Result, DeriveInput};
 
-/// Provides the hook to expand `#[derive(Index)]` into an implementation of `From`
+/// Provides the hook to expand `#[derive(DerefMut)]` into an implementation of `DerefMut`
 pub fn expand(input: &DeriveInput, trait_name: &'static str) -> Result<TokenStream> {
     let state = State::with_field_ignore_and_forward(
         input,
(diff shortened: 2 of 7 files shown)


JelteF:
Awesome work! I'll try to look at this soon

JelteF:
This is now released as 0.99.2

CGMossa:
I'm just about to put out an issue on why `derive(Display)` doesn't propogate "precision" or other formatting things. Maybe this PR is related to this reason?

PR #111, opened 2020-01-06
Fix description of example for From

birkenfeld:
With the addition of `ignore`, an impl for `u32` *is* generated and shown in the example.


JelteF:
This is now live on the docs website.

Issue #114, opened 2020-01-16
Why Deref/DerefMut without forward bounds T as Deref/DerefMut


kgv:
Hi.
This code:
```rust
#[derive(Deref, DerefMut)]
pub struct Has<T>(T)
where
    T: Clone + Send + Sync + 'static;
```

Expands to:
```rust
impl <T: ::core::ops::Deref> ::core::ops::Deref for Has<T> where T: Clone +
 Send + Sync + 'static {
    type
    Target
    =
    T;
    #[inline]
    fn deref(&self) -> &Self::Target { &self.0 }
}
impl <T: ::core::ops::DerefMut> ::core::ops::DerefMut for Has<T> where
 T: Clone + Send + Sync + 'static {
    #[inline]
    fn deref_mut(&mut self) -> &mut Self::Target { &mut self.0 }
}
```

But `T` need not be bounded `::core::ops::Deref`.

JelteF:
I don't understand what the problem is. Could you show the code that you would expect?

kgv:
@JelteF 
Change from `impl <T: ::core::ops::Deref>` to `impl <T>`

Expected:
```rust
impl <T> ::core::ops::Deref for Has<T>
where T: Clone + Send + Sync + 'static {
    type Target = T;
    #[inline]
    fn deref(&self) -> &Self::Target { &self.0 }
}
```

JelteF:
That seems like a bug indeed then. A PR would be much appreciated.

PR #115, opened 2020-01-20
Bound Deref/DerefMut only for forward

yankuu:


yankuu:
An attempt to fix #114.

JelteF:
I made some changes to where in the code this fix is actually implemented, but other than that it's the same. Thanks for the work on this.

JelteF:
Just released this as 0.99.3 as well. So you should be able to use the package from crates.io now.

PR #117, opened 2020-02-20
Fix the include glob for tests/

cuviper:


cuviper:
FWIW, you can check what is included before publishing with `cargo package --list`.

JelteF:
Ugh... Always those simple typos when you don't actually test if your change works

JelteF:
This is now released in version 0.99.5

PR #126, opened 2020-05-08
fix derivation of MulAssign for generic structs

rkuhn:
The code structure generated for AddAssign and MulAssign differs in what it supports: should we make AddAssign as permissive as MulAssign in that it accept a RHS as long as the underlying type’s operation will accept it?


JelteF:
Yes, this was a bug. The reason this slipped through was because `MulAssign` was not used in the `generics.rs` test file.

JelteF:
Thank you for your work on this. I just released this as v0.99.7

rkuhn:
Cool, thanks!

PR #129, opened 2020-05-30
Fix generic trait bound for AsRef, AsMut

matthew-mcallister:
The current version adds the unnecessary (and syntactically incorrect) trait bound `impl<T: ::std::convert::AsRef>`.

I did the smallest amount of refactoring necessary to make the change, but of course there is always room for more.


JelteF:
FYI this is now released as v0.99.8

PR #130, opened 2020-06-03
Make TryInto::Error return the original value instead of a str

ninegua:
When using the derived `TryInto/TryFrom` instances, if the result is an `Error`, the original value is consumed (if it is not a reference). This prevents the valid use case of chaining several `try_from` or `try_into` together, at least not without potentially expensive clones.

Returning a string error message is descriptive if the user program actually prints it, but IMHO that is not a very common use case because reliable programs should always handle errors. An additional reason against string error is that it is not friendly for localization/translation.

I could change the `Error` type to be a tuple that contains both the string & the original value if people really must have the string back.

### DIFF
--- a/doc/try_into.md
+++ b/doc/try_into.md
@@ -16,13 +16,16 @@ If that's not provided the default is `#[try_into(owned)]`.
 With `#[try_into]` or `#[try_into(ignore)]` it's possible to indicate which
 variants you want to derive `TryInto` for.
 
+In case of an error, the original value is not destructed but returned.
+This enables the chaining of several `try_from` (or `try_into`) calls.
+
 # Example usage
 
 ```rust
 # #[macro_use] extern crate derive_more;
 use core::convert::TryFrom;
 use core::convert::TryInto;
-#[derive(TryInto, Clone)]
+#[derive(TryInto, Clone, Eq, PartialEq, Debug)]
 #[try_into(owned, ref, ref_mut)]
 enum MixedData {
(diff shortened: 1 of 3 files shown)


JelteF:
Thanks for this PR, a few initial comments:
1. This can only be released as part of 1.0.0, since this is a breaking change.
2. Can you add a test that does the chaining?
3. The changed doctests do not compile. The easiest is usually to use [`cargo expand`](https://github.com/dtolnay/cargo-expand) to get the right code.

JelteF:
And I'm a bit undecided if I'd like to keep the user friendly error messages too. They are quite helpful in cases of a programming error.

ninegua:
One possibility is to support returning the original value through Err as an optional feature.

JelteF:
> One possibility is to support returning the original value through Err as an optional feature. This will not break backward compatibility.

I think this is good idea. Could you implement that? I don't think most people will require the original type to be returned. Something like:
1. `#[try_into(error_type(string))]`, this is the default
2. `#[try_into(error_type(input))]`, this would enable your desired behaviour

JelteF:
removing this from the 1.0 milestone as this doesn't need a breaking change

JelteF:
Added back 1.0 milestone, because I think returning an error by default that actually implements `Error` would be really good.

JelteF:
Closing this since there was no response from the original author, and I don't think the approach is the right one. let's continue to discuss in #130

Issue #99, opened 2019-11-02
Generalised deriving for enums


JelteF:
As was initially proposed in #87, it would be very useful to have a general way of deriving for enums. The basic idea is that when there's an enum we all variants forward to their fields. So in case of `Index` the following would work:
```rust
#[derive(Index)]
enum VecOrSlice<'slice> {
    Vec(Vec<123>)
    Slice(&'slice [123])
}
```

PR #134, opened 2020-06-27
Support enum forwarding for Index derive

JelteF:
This puts down the basis for fixing #99


JelteF:
Closing this as I'm not going to fix this. If anyone sees a need, feel free to use this code as an example.

Issue #119, opened 2020-03-05
make derive From more useful


xoac:
Hi guys

Minimal concept example
```rust
#[derive(From)]
enum MyEnum {
   #[from(i8,u8,i16,u16,i32,u32,i64)]
   Number(i64)
}
```

This would generate `From` trait for every type that can be converted to `i64` and it is listed in `from` attribute. 

What do you think?

JelteF:
What's the reason you would use this over the #[from(forward)] attribute?

xoac:
Oh I just read about use-case on enum. Sorry I missed it in doc but it's still useless in my case. 

Answering your question - this allow you to specify exactly what type you want to convert from. 

#[derive(From)]
```rust
enum Number {
  #[from(i8, i16, i32, i64)]
  Signed(i64)
  #[from(forward)] // this will not conflict with upper and bellow
  Unsigned(u64)
  // I don't wanna convert from integers just from floats
  #[from(f32, f64)]
  Float(f64)
}
```

The second reason is `forward` make it a bit out of control since you can't say from what types it will be generated (sometimes it's not what you want).

JelteF:
Hmm, yeah that's actually a good use case for this indeed. Feel free to make a PR for this implementing it. To make it more easy to stay backwards compatible though, please use the following syntax:
```
#[from(types(i8, i16, i32, i64))]
```

xoac:
I have tried to jump into code but it's a little bit complicated to me. I have just once worked with creating derive. Could you give me some general insight where I should start? It would be also helpful to add some doc-comments in `utils` module. Is there a way to chat with you if I would have concerns? 

Did you think about spiting this crate to sup-crates and then just reexport them I think people mostly are looking just for a single costume derive?:
- not all derive require so complicated utils to be implemented (and utils could be also be useful separate crate - not only for this project)
- it would be easier findable (in `Cargo.toml` you can specify only  5 [keywords](https://doc.rust-lang.org/cargo/reference/manifest.html#the-keywords-field))
- you would have probably more contributors

JelteF:
The code is indeed not the most clear and comments are quite lacking in most places. The way I would recommend starting with this is
1. create a separate test file, e.g. tests/from_types.rs.
2. Put in the syntax you want there
3. Use `cargo expand --test from_types` to see what comes out
4. Modify `from.rs` and `utils.rs` to make it output what you want
5. Repeat 3 and 4 until you have what you want.

[...]

PR #136, opened 2020-07-17
Allow additional explicit types for Into

tyranron:
Following from #135 

## What this PR does?

This PR implements [the suggested syntax](https://github.com/JelteF/derive_more/issues/119#issuecomment-595712232) for additional types conversions when deriving `Into`.

```rust
#[derive(Into)]
#[into(owned, ref, ref_mut)]
struct MyInt(u8);

#[derive(Into)]
#[into(owned(types(u16, u32, u64, u128)), ref, ref_mut, types(u8)]
struct Point2D(MyInt, MyInt)
```

## Design considerations

This PR changes a bit how additional types are collected in `utils`. Previously, they were collected into a single `HashSet`. Now, there is a separate `HashSet` for each `RefType` variant. This way we can allow the library user to choose different additional types for `owned`/`ref`/`ref_mut` impls.


tyranron:
ping @JelteF

tyranron:
ping @JelteF

tyranron:
ping @JelteF

JelteF:
I'm having a busy few months/weeks, so I'm not sure when I'll be able to review this. I'll try to make some time for it though.

src/utils.rs
```diff
@@ -1022,30 +1068,28 @@ fn parse_punctuated_nested_meta(
...
-                    return Err(Error::new(
-                        path.span(),
-                        format!(
-                            "Attribute doesn't support parameter `{}` here",
-                            quote! { #path }
-                        ),
-                    ));
-                };
+                let attr_name = path.get_ident().unwrap().to_string();
+                match (wrapper_name, attr_name.as_str()) {
```
JelteF, line 1072: This match is not equivalent to the if statement before. Some things will now error that didn't before, e.g. `not(ref)`. Although doesn't make much sense, but now it errors where it did not before. 

I guess this is on purpose, because this could now get confusing with something like `types(i32, i64), not(ref(types(i64)))`.

JelteF (approved):

src/utils.rs
tyranron: @JelteF yup, this is on purpose, because now we have more nesting layers for attribute arguments. So stricter validation here only helps.

JelteF:
Released as 0.99.10

PR #152, opened 2021-01-25
Allow complex additional types for From/Into

tyranron:
This PR extends `types()` attribute in`From`/`Into` derives to support types more complex than a simple ident.

```rust
#[derive(From, Into)]
#[from(types("Cow<'_, str>", "&str"))]
#[into(owned(types("Cow<'_, str>")))]
struct Name(String);
```


tyranron:
ping @JelteF

tyranron:
ping @JelteF

JelteF (approved):

PR #158, opened 2021-04-01
Update docs: Remove reference to MulSelf

rMazeiks:
If I understand correctly these have been removed in the latest version.


JelteF:
This is a good improvement indeed, but should also be done in `src/lib.rs`.

JelteF:
Thanks for this suggestion. I included this in the latest release and update to the docs.

PR #161, opened 2021-05-20
Add Unwrap derive

Monadic-Cat:
```rust
#[derive(::derive_more::Unwrap)]
enum Lol {
    Haha(i32),
    Hm,
}

fn main() {
    let a = Lol::Haha(10);
    println!("Results: {:?}", a.unwrap_haha())
}
```
prints `Results: 10`, and switching to `a.unwrap_hm()` prints this message:
> thread 'main' panicked at 'called `Lol::unwrap_hm()` on a `Haha` value', src/main.rs:18:33
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

### DIFF
--- a/src/unwrap.rs
+++ b/src/unwrap.rs
@@ -38,13 +38,18 @@ pub fn expand(input: &DeriveInput, trait_name: &'static str) -> Result<TokenStre
         let (data_pattern, ret_value, ret_type) = match variant.fields {
             Fields::Named(_) => panic!("cannot unwrap anonymous records"),
             Fields::Unnamed(ref fields) => {
-                let data_pattern = (0..fields.unnamed.len()).fold(vec![], |mut a, n| {
-                    a.push(format_ident!("field_{}", n));
-                    a
-                });
+                let data_pattern =
+                    (0..fields.unnamed.len()).fold(vec![], |mut a, n| {
+                        a.push(format_ident!("field_{}", n));
+                        a
+                    });
                 let ret_type = &fields.unnamed;
-                (quote! { (#(#data_pattern),*) }, quote! { (#(#data_pattern),*) }, quote! { (#ret_type) })
-            },
+                (
+                    quote! { (#(#data_pattern),*) },
+                    quote! { (#(#data_pattern),*) },
+                    quote! { (#ret_type) },
+                )
+            }
             Fields::Unit => (quote! {}, quote! { () }, quote! { () }),
         };
 
--- a/src/unwrap.rs
+++ b/src/unwrap.rs
@@ -48,13 +48,28 @@ pub fn expand(input: &DeriveInput, trait_name: &'static str) -> Result<TokenStre
             Fields::Unit => (quote! {}, quote! { () }, quote! { () }),
         };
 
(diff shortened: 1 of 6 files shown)


raffomania:
This is very nice. Please let me know if I can help :)

JelteF:
I think this feature sounds nice, but it needs some tests.

Monadic-Cat:
I can add docs for this some time in the next week or so, depending on when I'm free, and maybe I can do tests based on whatever `IsVariant` has? This is pretty much a copy and paste of that macro with a few tweaks.

Monadic-Cat:
While writing the docs based on `IsVariant`'s docs, I noticed and tested `#[unwrap(ignore)]`, which doesn't work. Its use of `.enabled_variant_data()` for some reason unconditionally reports all variants. It works with `IsVariant`, though, so I'm trying to pin down what's different between the two.

src/unwrap.rs
```diff
@@ -0,0 +1,87 @@
...
+use proc_macro2::TokenStream;
+use quote::{format_ident, quote};
+use syn::{DeriveInput, Fields, Ident, Result};
+
+pub fn expand(input: &DeriveInput, trait_name: &'static str) -> Result<TokenStream> {
+    let state = State::with_attr_params(
+        input,
+        trait_name,
+        quote!(),
+        String::from("unwrap_variant"),
```
JelteF, line 12: Should be `unwrap` to fix `#[unwrap(ignore)]`.

Monadic-Cat: That works, thank you!

Monadic-Cat:
I've now added documentation and tests based on what's there for the `IsVariant` derive, and fixed a, in retrospect, rather obvious bug.

Monadic-Cat:
Looking over CI failures now:

Do you have an MSRV of 1.36 (going based on the CI failure)?
If so, this feature can't use the `#[track_caller]` attribute in its generated code, which will make the panic messages less helpful.

Lemme `rustfmt` the new stuff right now...

JelteF:
> Do you have an MSRV of 1.36 (going based on the CI failure)?

Yes, so far that's the MSRV indeed. What version is the first one that supports `track_caller`?

I think you can probably do something similar to what we do for nightly detection: 
https://github.com/JelteF/derive_more/blob/d335514f9f732f2e1b45c4bc7f3af20502a87c5d/build.rs#L26-L35

i.e. create a feature called `track_caller` (or something) when the rust version supports `track_caller`. Then only add the `track_caller` attribute when the `track_caller` feature is enabled.

Monadic-Cat:
> Yes, so far that's the MSRV indeed. What version is the first one that supports `track_caller`?

That would be Rust [1.46](https://blog.rust-lang.org/2020/08/27/Rust-1.46.0.html#track_caller).

> I think you can probably do something similar to what we do for nightly detection:

Done!

JelteF:
The doctests are failing. To test them locally you need to use my fork of `cargo external-doc`. Which you can install like:
```
cargo install --git https://github.com/JelteF/cargo-external-doc --force
```

Monadic-Cat:
That looks like everything!

JelteF (approved):

JelteF:
This is included in v0.99.16

PR #172, opened 2021-09-26
Add documentation for DebugCustom derive

### LINKED ISSUE #68: derive custom Debug
Hello,

I enjoying using this crate to derive simple `Display` implementations, and it'd be swell if I could do the same for `Debug`.

Example use case, consider the classic `Point`:

```rust
#[derive(Display)]
#[display(fmt = "({}, {})", x, y)]
struct Point {
    x: i32,
    y: i32,
}
```

Frequently, I'll have a collection of `Point`s, and I want to debug that. I'm not going to derive `Display` on the collection, so I use `Debug`. However, for a structure as simple as `Point`, I'd rather always use the nicer format of `Display` (not the verbose derived `Debug`).

Thanks!

detly:
Per the comments in #68, this was missing from the docs.

I don't think the code in the `doc/` docs runs under `cargo test`, but I did check that my additions compile, run, and pass.

Closes #68

### DIFF
--- a/doc/display.md
+++ b/doc/display.md
@@ -33,6 +33,10 @@ i.e. `_0`, `_1`, `_2`, etc.
 The syntax does not change, but the name of the attribute is the snake case version of the trait.
 E.g. `Octal` -> `octal`, `Pointer` -> `pointer`, `UpperHex` -> `upper_hex`.
 
+A special case is the `DebugCustom` trait, which derives a `Debug` implementation instead of a `Display` implementation.
+The attribute for formatting is named `#[debug(..)]` when using `DebugCustom`.
+The arguments are the same as for `#[display(..)]`.
+
 # Generic data types
 
 When deriving `Display` (or other formatting trait) for a generic struct/enum, all generic type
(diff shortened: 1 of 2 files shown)


doc/display.md
```diff
@@ -105,6 +109,17 @@ use std::path::PathBuf;
...
+#[debug(fmt = "MyIntDbg(as hex: {:x}, as dec: {})", _0, _0)]
+struct MyIntDbg(i32);
+
+#[derive(Display)]
+#[display(fmt = "({}, {})", x, y)]
+struct Point2D {
+    x: i32,
+    y: i32,
+}
+
```
JelteF, line 122: This is duplicated
```suggestion
```

JelteF (commented):
Thank you for this PR!

> I don't think the code in the doc/ docs runs under cargo test

I'm using a fork of `cargo-external-doc` to run the doctests in the `doc/` directory. You can install it with:
```
cargo install --git https://github.com/JelteF/cargo-external-doc
```

doc/display.md
detly: Ooops.

JelteF (approved):

PR #175, opened 2021-11-10
Move MSRV test to separate job with `cargo +nightly update -Z minimal-versions`

ilslv:
This PR moves MSRV test into a separate job and adds `cargo +nightly update -Z minimal-versions` to downgrade into lowest compatible version from `Cargo.toml`. 

For now, this crate depends on `syn v1.0.0` and higher, which isn't quite true, because `Path::get_ident()` method seen [here][1] wasn't introduced until [`syn v1.0.3`][2]. To avoid problems like this, I've fixed main dependencies version to latest stable ones. 

[1]: https://github.com/JelteF/derive_more/blob/master/src/utils.rs#L995
[2]: https://github.com/dtolnay/syn/releases/tag/1.0.3


tyranron (dismissed):
Run CI

ilslv:
@tyranron bumping `rustc_version` to `0.4.0` works locally for me

tyranron (changes requested):
@ilslv so, at this moment `-Z minimal-versions` doesn't work because of `peg 0.5` in `[build-dependencies]` not working correctly with `-Z minimal-versions`.

Actually there is a `peg 0.7` already, but I doubt @JelteF will agree migrate to it, because at the time the whole idea of introducing `peg` was to avoid additional `[dependencies]` for parsing, while starting from `peg 0.6` it works like `[dependencies]` not the `[build-dependencies]`.

Is there any alternatives to `peg` which can produce parser in build time without living in direct `[dependencies]`, while are maintained?

tyranron:
@ilslv 

> bumping `rustc_version` to `0.4.0` fixes MSRV locally for me

Oh, nice! ♥️

So we can postpone `peg` question for now.

tyranron:
@ilslv now it doesn't work because 1.36 Rust doesn't understand well newer `.lock` file format. 🙈

I guess we need for now to separate check for `-Z minimal-versions` and MSRV in different jobs, until MSRV is bumped up.

tyranron (approved):
@JelteF to summaize this PR:
- adds CI job checking whether crate works OK on minimal crates versions described in `Cargo.toml`
- fixes Clippy lints on latest nightly

tyranron:
@JelteF can we hope for a patch release with this, so downstream crates won't fail on `-Z minimal-versions` check?

JelteF:
Released v0.99.17 just now, which includes this change.

PR #176, opened 2021-12-03
Add support for #derive[FromStr] for enums

### LINKED ISSUE #59: Implement FromStr for Enum
I have an enum with a long list like this:

```rust
enum ThingType {
    A,
    B,
    ...
    Z,
}
```

I want to be able to do

```rust
ThingType::from_str("A")
```
and get `ThingType::A`.

aj-bagwell:
This allows deriving FromStr for enums with no fields in the obvious way

fixes #59

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,7 +29,7 @@ proc-macro = true
 [dependencies]
 proc-macro2 = "1.0"
 quote = "1.0"
-syn = "1.0.3"
+syn = "1.0.81"
 convert_case = { version = "0.4", optional = true}
 
 [build-dependencies]
@@ -51,7 +51,7 @@ deref_mut = []
 display = ["syn/extra-traits"]
 error = ["syn/extra-traits"]
 from = ["syn/extra-traits"]
-from_str = []
+from_str = ["convert_case"]
 index = []
 index_mut = []
 into = ["syn/extra-traits"]
--- a/src/from_str.rs
+++ b/src/from_str.rs
@@ -1,6 +1,8 @@
+use crate::utils::{DeriveType, HashMap};
 use crate::utils::{SingleFieldData, State};
+use convert_case::{Case, Casing};
 use proc_macro2::TokenStream;
-use quote::quote;
+use quote::{format_ident, quote};
 use syn::{parse::Result, DeriveInput};
 
 /// Provides the hook to expand `#[derive(FromStr)]` into an implementation of `FromStr`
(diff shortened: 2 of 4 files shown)


src/from_str.rs
```diff
@@ -41,7 +52,81 @@ pub fn expand(input: &DeriveInput, trait_name: &'static str) -> Result<TokenStre
...
+        }
+
+        impl std::error::Error for #err_name {}
+
+        impl #trait_path for #input_type
+        {
+            type Err = #err_name;
+            #[inline]
+            fn from_str(src: &str) -> ::core::result::Result<Self, Self::Err> {
+                Ok(match src.to_lowercase().as_str() {
```
aj-bagwell, line 123: I considered adding `.replace(&[' ', '_', '-'], "")` here so that it would mach against kebab-case and SNAKE_CASE too.

src/from_str.rs
```diff
@@ -41,7 +52,81 @@ pub fn expand(input: &DeriveInput, trait_name: &'static str) -> Result<TokenStre
...
+            }
+        }
+    }
+
+    let trait_path = state.trait_path;
+
+    quote! {
+
+        #[derive(Debug, Clone, PartialEq, Eq)]
+        #visibility struct #err_name;
```
aj-bagwell, line 108: I'm not sure I like creating a type out of nothing here, but I did not like the other options better:

- could use `&'static str` but that does not impelement `std::error::Error` which make it annoying for combining with other errors
- could have a `derive_more::ParseError` but that makes the dependencies complicates as you cant have code and procmarcos in the same dependency

aj-bagwell:
Forced push changes that hopefull fix the failing tests

Cargo.toml
```diff
@@ -51,7 +51,7 @@ deref_mut = []
 display = ["syn/extra-traits"]
 error = ["syn/extra-traits"]
 from = ["syn/extra-traits"]
-from_str = []
+from_str = ["convert_case"]
```
JelteF, line 54: This should fix the current compilation issue in the feature tests.
```suggestion
from_str = ["convert_case", "syn/extra-traits"]
```

aj-bagwell: Thanks, I just furoced pushed a change that bumps `syn` to 1.0.81 (which is the version in the lock file) and changed the code to use .is_empty() instead which means it does not need the extra traits

aj-bagwell: Bah looks like I also need to ditch the `x == from_str(x.to_string())` test as I can't `#derive[Display]` if the display feature is not set.

aj-bagwell:
The test are all now passing and I'm happy with my commits as they stand, @JelteF let me know if there are any more changes that you want me to make.

src/from_str.rs
```diff
@@ -41,7 +51,81 @@ pub fn expand(input: &DeriveInput, trait_name: &'static str) -> Result<TokenStre
...
+            cases.push(quote! {
+                #canonical => #input_type::#variant,
+            })
+        } else {
+            for variant in variants {
+                let variant_str = variant.to_string();
+                cases.push(quote! {
+                    #canonical if(src == #variant_str) => #input_type::#variant,
+                })
+            }
```
JelteF, line 98: this edge case is missing a test

JelteF (dismissed):
I think this is good except for a missing test for the smart case sensitive behaviour.

aj-bagwell:
I have added `enum_test_case_sensitive` to check if falls back to case sentive matching when it matters and rebased onto master to bring it up to date

JelteF:
Thanks a lot for your work on this!

PR #179, opened 2021-12-21
Add doc comments for Unwrap, IsVariant

### LINKED ISSUE #159: #[derive(IsVariant)] clashes with #![deny(missing_docs)]
Trying to run the following code

```rust
#![deny(missing_docs)]
//! Documented crate.
use derive_more::IsVariant;

/// Documented enum.
#[derive(IsVariant)]
pub enum Documented{
    /// Documented variant
    Foo,
    /// Documented variant
    Bar,
}
```

will result in two errors, 

> missing documentation for an associated function rustc(missing_docs) [6,10]
> missing documentation for an associated function rustc(missing_docs) [6,10]

I assume that's a bug on the side of derive_more? At the moment my only workaround is to #![allow(missing_docs)] the entire module, which is a bit unfortunate.

(This is my first filed issue on GitHub btw, I hope I did this right 😳 )

Technohacker:
Prevents a linting issue from `missing_docs` on the derived functions

Fixes #159


JelteF:
It seems the new code does not compile on 1.36 because of this issue https://github.com/rust-lang/rust/issues/52607
The second comment in the issue describes a workaround that can probably be used. This bug was only fixed in 1.50

Can you check if it's possible to use this workaround?

Technohacker:
Hmm, I'll have a check on that

Technohacker:
As an aside, would it make sense to add something like `unwrap_*_ref` and `unwrap_*_ref_mut` functions to the Unwrap derive, or maybe make a new derive macro for these? Intended for something similar to the enum_as_inner crate, but without having to unwrap the options in between when the type of enum is known

Technohacker:
Turns out my previous code would have expanded the `stringify!` macro after the Unwrap/IsVariant derive macros, rather than expanding it within the derive itself. New change should fix that

PR #181, opened 2022-01-19
Replace `peg` crate with manually implemented parser

ilslv:
In `0.6` release of `peg` [`build.rs` parser generation support was removed](https://github.com/kevinmehall/rust-peg/releases/tag/0.6.0), so to keep our dependencies up to date, we've decided to re-implement parser. Design is heavily inspired by [`nom`](https://docs.rs/nom/latest/nom/) crate.

This PR also fixes some inconsistencies with original `format_string` syntax and adds support for non-ascii identifiers. To make this possible I've added 1 dependency: `unicode-xid`, but actually `syn` crate already depends on it.

Also parser supports non-integer parameters, because in follow-up PR I'm planning to add support for interpolated strings [introduces in `1.58` release](https://blog.rust-lang.org/2022/01/13/Rust-1.58.0.html#captured-identifiers-in-format-strings) while preserving MSRV.


tyranron (approved):
@ilslv LGTM, thanks! 🍻

@JelteF I'd like you to look at this as well, whether you're OK with these changes.

tyranron:
ping @JelteF

JelteF (approved):
I haven't looked in detail at the new parser code, but it looks much more understandable than the previous auto generated one. Given none of our display tests fail, I think it's fine to merge.

PR #182, opened 2022-01-20
Add support for captured identifiers

ilslv:
Since Rust 1.58 `format_args!` macro can capture identifiers: `format!("Prefix: {some_local_variable}")`. This PR adds support for this feature while preserving MSRV by transforming `format!("Prefix: {field}")` into `format!("Prefix: {field}", field = field)`.

### Limitations

This approach doesn't work well with unnamed fields that are aliased with `_0, _1, ...` because [until 1.41 identifiers beginning with underscore weren't supported](https://github.com/rust-lang/rust/pull/66847)


JelteF:
> while preserving MSRV by transforming format!("Prefix: {field}") into format!("Prefix: {field}", field = field).

I haven't looked at the code yet, but if it makes the implementation easier I wouldn't mind if it this new syntax is only supported on 1.58+, as long as the old syntax still works on the current MSRV.

ilslv:
> if it makes the implementation easier I wouldn't mind if it this new syntax is only supported on 1.58+, as long as the old syntax still works on the current MSRV.

Both the new syntax and the old one work fine on MSRV [with 1 limitation described in PR message](https://github.com/JelteF/derive_more/pull/182#issue-1109017302). 

> I haven't looked at the code yet

This PR depends on a new parser implemented in https://github.com/JelteF/derive_more/pull/181 so we should wait until it's merged.

JelteF:
I merged #181. Please update this PR and let @tyranron review it.

tyranron (approved):
@ilslv LGTM, thanks!

PR #185, opened 2022-02-25
Error on invalid `#[display(...)]` arguments

### LINKED ISSUE #184: Invalid syntax in Display fmt fails silently
Consider this code:

```rust
use derive_more::Display;

#[derive(Display)]
#[display(fmt = "Stuff({})", .bar)]
pub struct Foo {
    bar: String
}

fn main() {
    println!("{}", Foo { bar: "test".into() });
}
```

This prints just `test`. Similarly, if you do this, it still prints `test`: 

```rust
#[derive(Display)]
#[display(fmt = "Stuff({})", .barrrrrrr)]
pub struct Foo {
    bar: String
}
```

The correct syntax here would be to use just `bar` in the format, but that's not super obvious since no error is raised.

Perhaps this could be fixed to inform the user that the format is just wrong?

Thanks!

ilslv:
Resolves #184

This PR errors on invalid `#[display(...)]` arguments and adds tests for compiler errors with [`trybuild`](https://docs.rs/trybuild/1.0.56/trybuild/) crate.


ilslv:
## FCM

```
Error on invalid `#[display(...)]` arguments (#185, #184)
```

ilslv:
@tyranron ping

PR #190, opened 2022-05-26
`derive(Default)` on enum

ModProg:
I know that there will be a `derive(Default)` for enums in the `std` soon, but AFAICT that only implements default for unit variants.

Due to name collisions, this has to be called `EnumDefault`. This currently has neither tests nor documentation, but I'll add those when you'd approve of this addition

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -49,6 +49,7 @@ add = []
 as_mut = []
 as_ref = []
 constructor = []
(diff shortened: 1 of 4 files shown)


ModProg:
> @ModProg have you considered https://docs.rs/smart-default for this?

Right, maybe it makes sense not to implement the same thing twice :)

PR #192, opened 2022-06-09
display: enum match arms: use 'Self' (to fix incompatibility with `#![deny(clippy::use_self)]`

phip1611:
This PR makes the generated code compatible with a global setting of  `#![deny(clippy::use_self)]` by using `Self::Variant` instead of `EnumName::Variant`.

### DIFF
--- a/src/display.rs
+++ b/src/display.rs
@@ -503,9 +503,8 @@ impl<'a, 'b> State<'a, 'b> {
                             } else {
                                 self.infer_fmt(&v.fields, &v.ident)?
                             };
-                            let name = &self.input.ident;
                             let v_name = &v.ident;
-                            Ok(quote_spanned!(fmt.span()=> #arms #name::#v_name #matcher => write!(
+                            Ok(quote_spanned!(fmt.span()=> #arms Self::#v_name #matcher => write!(
                                 _derive_more_display_formatter,
                                 #outer_fmt,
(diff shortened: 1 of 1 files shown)


tyranron:
@phip1611 well, this change doesn't survive the MSRV check.

@JelteF are we OK to bump up MSRV to at least 1.51 (last 10 versions)? Or do you see some sense to keep MSRV such outdated?
Additionally, I'd like to place `#[automatically_derived]` attribute on the every generated code, so the style linters may omit the generated code (doesn't solve this issue due to the [Clippy bug](https://github.com/rust-lang/rust-clippy/issues/6902), however).

phip1611:
btw, I fixed the latest (unrelated) clippy error as well.

JelteF:
I am not opposed to bumping MSRV to something less ancient, but I would like to do that only when increasing the major version of the library. Marking this lib as 1.0 is way overdue anyway. 

I cleaned up the 1.0.0 milestone to only contain things that require breaking changes, it turned out that was only a single thing.

tyranron:
@JelteF in case we don't want to do `1.x` releases yet, we always may release just `0.100.x` one, containing the necessary breaking changes.

phip1611:
any update here? @tyranron  @JelteF

tyranron:
@phip1611 bumping MSRV is OK, as we're going to release 1.0.0.

phip1611:
I rebased the MR onto the latest master. Can you approve the CI, please?

PR #200, opened 2022-08-31
Replace `fn backtrace()` with `Provider` API in `Error` derive

### LINKED ISSUE #195: backtrace method was removed by Rust

ilslv:
Fixes #195 similar to https://github.com/dtolnay/thiserror/pull/182

### DIFF
--- a/doc/error.md
+++ b/doc/error.md
@@ -50,7 +50,7 @@ ignored for one of these methods by using `#[error(not(backtrace))]` or
 #![feature(error_generic_member_access, provide_any)]
 # #[macro_use] extern crate derive_more;
 # use std::error::Error as _;
-use std::backtrace::Backtrace;
+use std::{any, backtrace::Backtrace};
 
 // std::error::Error requires std::fmt::Debug and std::fmt::Display,
 // so we can also use derive_more::Display for fully declarative
@@ -90,8 +90,22 @@ enum CompoundError {
     WithSource {
         source: Simple,
     },
+    #[from(ignore)]
+    WithBacktraceFromSource {
+        #[error(backtrace)]
+        source: Simple,
+    },
+    WithDifferentBacktrace {
+        source: Simple,
+        backtrace: Backtrace,
+    },
     WithExplicitSource {
         #[error(source)]
+        explicit_source: WithSource,
+    },
+    #[from(ignore)]
+    WithBacktraceFromExplicitSource {
+        #[error(backtrace, source)]
         explicit_source: WithSource,
     },
     Tuple(WithExplicitSource),
@@ -100,7 +114,7 @@ enum CompoundError {
 
 fn main() {
     assert!(Simple.source().is_none());
-    assert!(Simple.backtrace().is_none());
+    assert!(any::request_ref::<Backtrace>(&Simple).is_none());
     assert!(WithSource::default().source().is_some());
     assert!(WithExplicitSource::default().source().is_some());
     assert!(Tuple::default().source().is_some());
@@ -110,7 +124,7 @@ fn main() {
         backtrace: Backtrace::capture(),
     };
     assert!(with_source_and_backtrace.source().is_some());
-    assert!(with_source_and_backtrace.backtrace().is_some());
+    assert!(any::request_ref::<Backtrace>(&with_source_and_backtrace).is_some());
 
     assert!(CompoundError::Simple.source().is_none());
     assert!(CompoundError::from(Simple).source().is_some());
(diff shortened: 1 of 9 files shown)


tyranron (approved):

tyranron:
@JelteF would you be so kind to move requirement from `Test Rust 1.36.0` CI jobs to `Test Rust 1.56.0` in `master` branch protection settings?

tyranron (changes requested):
@ilslv hmm...regarding the CI failure. I thought it was because of my changes, but I can reproduce the same test failures localy on your last commit in `2022-09-08` nightly Rust. Could you look into it?

tests/error/nightly/mod.rs
```diff
@@ -86,6 +86,8 @@ impl Default for BacktraceErr {
 
 impl Error for BacktraceErr {
     fn provide<'a>(&'a self, demand: &mut std::any::Demand<'a>) {
-        demand.provide_ref(&self.backtrace).provide_value(|| 42);
+        demand
+            .provide_ref::<Backtrace>(&self.backtrace)
+            .provide_value::<i32>(42);
```
ilslv, line 91: @tyranron downside of using nightly is that [API can change at any time][1] 🙃 

[1]: https://github.com/rust-lang/rust/pull/99583/files#diff-0752889661748b8a15a597d7156127b6fb90fdeda0627be50e07f3f785bd0f4dR799

tyranron: @ilslv let's see for a while how it will play.

tyranron (approved):

doc/error.md
```diff
@@ -90,17 +90,32 @@ enum CompoundError {
...
+        #[error(backtrace, source)]
+        explicit_source: WithSource,
+    },
     Tuple(WithExplicitSource),
     WithoutSource(#[error(not(source))] Tuple),
 }
 
 fn main() {
     assert!(Simple.source().is_none());
-    assert!(Simple.backtrace().is_none());
```
JelteF, line 103: The doc text above still mentions the backtrace method. This should be changed to talk about this `any::request_ref::<Backtrace>()` function

src/error.rs
```diff
@@ -203,16 +199,67 @@ impl<'input, 'state> ParsedFields<'input, 'state> {
...
+            }
+        });
+        let backtrace_provider = self
+            .source
+            .filter(|source| *source == backtrace)
+            .is_none()
+            .then(|| {
+                let backtrace_expr = &self.data.members[backtrace];
+                quote! {
+                    demand.provide_ref::<std::backtrace::Backtrace>(&#backtrace_expr);
```
JelteF, line 218: Food for thought: One of the reasons for this provider API (afaict) is so more types can be provided that just Backtrace, without needing to add more methods to the `Error` trait. 

So probably we want to allow this too, e.g. to embed some additionaly context such as trace IDs in there for distributed tracing. This would probabyl be better as a separate PR. I think it makes sense to keep dedicated and easy to use support for the `Backtrace` type.

Another idea is to also provide the `source` through the provider API, like is done in the rust docs

```rust
impl std::error::Error for Error {
    fn provide<'a>(&'a self, demand: &mut Demand<'a>) {
        demand
            .provide_ref::<MyBacktrace>(&self.backtrace)
            .provide_ref::<dyn std::error::Error + 'static>(&self.source);
    }
}
```
Source: https://doc.rust-lang.org/nightly/std/error/trait.Error.html

JelteF (changes requested):
Looks good overall. but docs need to be updated. Thanks for all the work on this.

src/error.rs
ilslv: @JelteF I've already [raised question of what should be provided][1] and this implementation mirrors [`thiserror`][2]. As `Provider` API is quite new and `source()` default impl still returns `None`, I think that we shouldn't rush with providing `dyn Error` and wait how `Provider` API will be used throughout the ecosystem. 




[1]: https://github.com/JelteF/derive_more/issues/195#issuecomment-1231616625
[2]: https://github.com/dtolnay/thiserror/pull/182

tyranron:
@JelteF ping

JelteF (approved):

Issue #186, opened 2022-02-27
Complete enum derives with getters which return Option<&T>


MTCoster:
*I'm submitting this as an issue to gather feedback; I'm happy to do the work on an implementation if this is a desired feature.*

---

Currently I can use a combination of `IsVariant` and `Unwrap` to build an `Option<T>` for a specific variant of an enum.

It would be nice to be able to derive a getter which combines this functionality. For instance:

```rust
#[derive(GetVariant)]
enum MyEnum<A, B> {
    VarA(A),
    VarB(B),
}
```

Becomes:

```rust
enum MyEnum<A, B> {
    VarA(A),
    VarB(B),
}

impl <A, B> MyEnum<A, B> {
    fn var_a(&self) -> Option<&A> {
        match self {
            Self::VarA(var_a) => Some(var_a),
            _ => None,
        }
    }

    fn var_b(&self) -> Option<&B> {
        match self {
            Self::VarB(var_b) => Some(var_b),
            _ => None,
        }
    }
}
```

This somewhat mimics the implementation of `.ok()` and `.err()` on std's `Result<T, E>`, but those return `Option<T>` rather than `Option<&T>`. In that instance, this distinction doesn't matter as first convert to a `Result<&T, &E>` with `.as_ref()`.

JelteF:
This si now possible using `#[try_unwrap(ref)]`: https://docs.rs/derive_more/1.0.0-beta.6/derive_more/derive.TryUnwrap.html

PR #206, opened 2022-10-08
Add try/ref versions to unwrap

rassvet2:
I extended Unwrap to add one that returns Option, one that returns reference, and a combination of the two, for my use.
It's ready to merge, but I don't think it's well discussed and considered. `try_unwrap_<variant>_ref` is... little uncool.
So I'll leave it as a draft to solicit opinions.

```rust
#[derive(Unwrap)]
enum Number {
    Integer(i64),
    Decimal(f64),
    #[unwrap(ref)] // this affects only `FromString`
    FromString(String),
}

#[derive(Unwrap)]
#[unwrap(ref)] // this affects the entire enum,
               // and adds `unwrap_<name>_ref` and `try_unwrap_<name>_ref`.
enum Maybe<T> {
    Just(T),
    Nothing,
}

fn main() {
    assert_eq!(Maybe::Just(1).unwrap_just(), 1);
    assert_eq!(Maybe::Just(2).try_unwrap_just(), Some(2));
    assert_eq!(Maybe::Nothing.try_unwrap_just(), None);
    assert_eq!(*(&Maybe::Just(42)).unwrap_just_ref(), 42);
}
```

This relates to #186.
`var_a` looks better than `try_unwrap_var_a_ref`. But I think in that case it probably should not be implemented in Unwrap.
Initially, I added it in Unwrap because it was tedious, but maybe it should be split into TryUnwrap, UnwrapRef, and TryUnwrapRef.

Is there anything else I have to do to be merged?


rassvet2:
There is a known issue with *_ref naming conflicts when an enum has SomeVariant and SomeVariantRef.
I need advice on this as well.

JelteF:
I didn't look at the code in detail, but a few thoughts on the general design:
1. For completeness this should also add `unwrap_xxx_mut` versions (unless somehow hard to do) 
2. I really don't like the `try_unwrap` prefix. Some other options I can think of: `as_thing`, `try_into_thing`, `try_thing`. I think I like `try_into_thing` the best. (I will continue calling it `try_unwrap` in the rest of this comment. 
3. Should `try_unwrap` return an error instead?


This seems rare enough that i wouldn't consider it a problem:
> There is a known issue with *_ref naming conflicts when an enum has SomeVariant and SomeVariantRef.

JelteF:
To clarify: The most open questions and discussion points for me are around the `try_unwrap` version of this derive. I think it makes sense to have the `unwrap_thing_ref` and `unwrap_thing_mut` without any further discussion on the design.

rassvet2:
for 1 and 2:

Yes, I will. I like the shortness of `as`, but it seems misleading. Looking at std, it seems that functions starting with `as_` rarely return `Option` or `Result`. So I'm going to use `try_into`.

for 3 and 4:

Makes sense. I just forgot that `TryInto` returns `Result`, not `Option`. I will make it returns the original value as an error for now.
As for `try_unwrap`, I'm going to delete it temporarily and separate it into a new PR.

JelteF:
@rassvet2 are you still planning on making your suggested changes?

rassvet2:
I'm sorry I left this hanging for so long.
I will work on this. It may take about two weeks.

rassvet2:
It's ready for review. Finally, I decided like:
- `try_unwrap_*` was renamed to `try_into_*` and moved to `TryIntoVariant` (Isn't it a bit long compared to Unwrap?)
- Both `Unwrap` and `TryIntoVariant` basically generate `owned` versions only, but `ref` and `ref_mut` can also be generated by specifying them in the attribute.
- `try_into_*` returns a `Result` value. This will be `Ok` if the conversion succeeds, and `Err` if it fails. `Ok` contains the converted value, and `Err` contains the original value.
  This makes it useful when you do not want to lose ownership if conversion fails.

doc/unwrap.md
```diff
@@ -1,20 +1,30 @@
...
 }
+
+fn main() {
+    assert_eq!(Maybe::Just(1).unwrap_just(), 1);
+
+    // Panics if variants are different
+    // assert_eq!(Maybe::Nothing.unwrap_just(), /* panic */);
+    // assert_eq!(Maybe::Just(2).unwrap_nothing(), /* panic */);
+
+    assert_eq!(*(&Maybe::Just(42)).unwrap_just_ref(), 42);
```
JelteF, line 26: The instant dereference looks a bit weird in the example. This looks like more common code I think:
```suggestion
    assert_eq!(&Maybe::Just(42)).unwrap_just_ref(), &42);
```

src/try_into_variant.rs
```diff
@@ -0,0 +1,141 @@
...
+        let doc_mut = format!(
+            "Attempts to convert this mutable reference to the `{enum_name}::{variant_ident}` variant.\n",
+        );
+        let doc_else = "Returns the original value if this value is of any other type.";
+        let func = quote! {
+            #[inline]
+            #[track_caller]
+            #[doc = #doc_owned]
+            #[doc = #doc_else]
+            pub fn #fn_name(self) -> Result<(#(#data_types),*), Self> {
```
JelteF, line 72: Instead of returning `Self` let's do the same thing we now do for `TryInto` where we return a custom error type: https://github.com/JelteF/derive_more/blob/master/src/convert.rs

JelteF: You might need to create a `TryIntoVariantError` type if the `TryIntoError` doesn't have the right fields.

JelteF (changes requested):
Overall this looks good! But it needs a rebase with the current master branch. Quite a few things have changed in the file organization of the library since this PR was originally opened.

src/try_into_variant.rs
rassvet2: I'm going to create `TryIntoVariantError`.
It contains the input value and variant names of the input/output.
And it'll be displayed like `Attempted convertion to Enum::VariantB on a Enum::VariantA value`.
I'm also thinking of implementing a Deref trait to get the input field. (It seems like an antipattern tho) What do you think?

rassvet2: Honestly, I'm not good at English. So please also review the documents and panic messages.

rassvet2:
I'm wondering again if I should rename `TryIntoVariant`. Now I don't feel it's a good name. Since these functions don't convert one enum variant into another, they're not a conversion like `std::convert`. In that respect, `Unwrap` was a good name as it "unwraps" an enum variant into a tuple of its fields.
However, I can't think of anything that fits. Do you agree with `TryIntoVariant`, or not?

JelteF:
>  Since these functions don't convert one enum variant into another, they're not a conversion like std::convert

That's a really good point. I think you're right that TryUnwrap is probably the best name. Both `Rc` and `Arc` have a `try_unwrap` method too that does exactly this, so there's also prior art from `std` on this naming. I'm sorry for the back and forth on the naming.

src/convert.rs
```diff
@@ -45,3 +45,52 @@ impl<T> fmt::Display for TryIntoError<T> {
 
 #[cfg(feature = "std")]
 impl<T: fmt::Debug> std::error::Error for TryIntoError<T> {}
+
+/// Error returned by the derived [`TryIntoVariant`] implementation.
+///
+/// [`TryIntoVariant`]: macro@crate::TryIntoVariant
+#[derive(Clone, Copy, Debug)]
+pub struct TryIntoVariantError<T> {
```
JelteF, line 53: Let's move this to a dedicated file `try_unwrap.rs` file

impl/doc/try_unwrap.md
```diff
@@ -0,0 +1,80 @@
...
+# enum Maybe<T> {
+#     Just(T),
+#     Nothing,
+# }
+#
+impl<T> Maybe<T> {
+    pub fn try_unwrap_nothing(self) -> Result<(), TryUnwrapError<Self>> {
+        match self {
+            Maybe::Nothing => Ok(()),
+            val @ _ => Err(panic!("stub")),
```
JelteF, line 58: This example of the `Err` contents seems incorrect/outdated.

tests/try_unwrap.rs
```diff
@@ -0,0 +1,95 @@
...
+    );
+    assert_eq!(
+        (&mut Maybe::Just(42))
+            .try_unwrap_nothing_mut()
+            .map_err(|e| e.to_string()),
+        Err(
+            "Attempt to call `Maybe::try_unwrap_nothing_mut()` on a `Maybe::Just` value"
+                .into()
+        ),
+    );
```
JelteF, line 94: Should have some test that checks that the old value can be extracted from the `Err`.

JelteF (dismissed):
Looks pretty good overal, just some small suggestions. Also there's a small merge conflict in the CHANGELOG.md now that #256 is merged which cleaned it up a bit.

impl/doc/try_unwrap.md
rassvet2: I didn't think this place needed to be the same as the actual code, so I placed a placeholder instead.
It doesn't panic in actuality, but the code to create a new instance of error is not short.
I did this to keep the sample code simple.

JelteF: Okay that's reasonable, the real code is indeed quite a bit of an implementation detail. Let's make the intent a bit clearer though then. Something like this.
```suggestion
            val @ _ => Err(panic!("This actually returns a TryUnwrapError containing self, panic is used here to keep sample simple")),
```

JelteF: Something shorter like this also fine with maybe a more descriptive comment at the top. I think another macro than `panic!` would also be clearer, since the normal `unwrap` actually does panic.
```suggestion
            val @ _ => Err(todo!("create TryUnwrapError here")),
```

rassvet2: How about this?

```suggestion
            val @ _ => Err(todo!("TryUnwrapError::new(val, /* omitted */)")),
```

JelteF: Looks perfect

JelteF (approved):

PR #209, opened 2022-10-13
Rework external docs (#207)

### LINKED ISSUE #207: More meaningful docs on docs.rs
I saw that derive_more has been doing its own thing on jeltef.github.io [since at least March 2021](https://github.com/JelteF/derive_more/commit/1e839c4f695460b4dacaa6f3b35c2cd5d792bac8), and the docs at https://docs.rs/derive_more/0.99.17 do not include any of the content from https://github.com/JelteF/derive_more/tree/v0.99.17/doc.

Since Rust 1.54 (July 2021) rustdoc knows how to include external markdown files as docs. See https://blog.rust-lang.org/2021/07/29/Rust-1.54.0.html#attributes-can-invoke-function-like-macros. Could you consider whether this means the bespoke website can be dropped in favor of docs.rs at this point?

tyranron:
Resolves #207  

This PR uses `#[doc = include_str!(...)]` introduced in Rust 1.54, to embed `README.md` and `doc/*.md` files directly into the generated Rust documentation without a need of any other external tool.

- [x] Embed docs via `#[doc = include_str!(...)]`.
- [x] Preserve working links in `README.md` on GitHub (currently they refer `doc/*.md` files on GitHub, but after first release may be easily changed to `docs.rs`).
- [x] Switch tests of external docs to 2021 Rust edition and fix formatting problems.
- [x] Use `feature(doc_auto_cfg)` for docs on https://docs.rs
- [x] Remove redundant toolchain regarding Rust docs generation.


tyranron:
@JelteF after merging this PR, I do plan to refresh CI pipeline in another one.

README.md
```diff
@@ -140,43 +152,43 @@ features = ["from", "add", "iterator"]
...
 // use the derives that you want in the file
 use derive_more::{Add, Display, From};
 ```
 If you're still using Rust 2015, add this instead:
-```rust
+```rust,edition2015
 extern crate core;
 #[macro_use]
 extern crate derive_more;
+# fn main() {} // omit wrapping statements above into `main()` in tests
```
JelteF, line 163: This line is not hidden in github its rendering of the readme. Which is the main reason I copied lib.rs to the readme all the time. I don't think there's a way around this sadly.

README.md
```diff
@@ -140,43 +152,43 @@ features = ["from", "add", "iterator"]
...
-[`Sum`-like]: https://jeltef.github.io/derive_more/derive_more/sum.html
-[`IndexMut`]: https://jeltef.github.io/derive_more/derive_more/index_mut.html
-[`DerefMut`]: https://jeltef.github.io/derive_more/derive_more/deref_mut.html
-[`AddAssign`-like]: https://jeltef.github.io/derive_more/derive_more/add_assign.html
-[`MulAssign`-like]: https://jeltef.github.io/derive_more/derive_more/mul_assign.html
-
-[`Constructor`]: https://jeltef.github.io/derive_more/derive_more/constructor.html
-[`IsVariant`]: https://jeltef.github.io/derive_more/derive_more/is_variant.html
-[`Unwrap`]: https://jeltef.github.io/derive_more/derive_more/unwrap.html
+[`From`]: doc/from.md
```
JelteF, line 169: These links should point to the docs.rs link instead. Otherwise the readme contains broken links. I think this is another good reason to keep the README and lib.rs docs different (although very similar).

src/lib.rs
```diff
@@ -270,7 +115,7 @@ macro_rules! create_derive(
     ($feature:literal, $mod_:ident, $trait_:ident, $fn_name: ident $(,$attribute:ident)* $(,)?) => {
         #[cfg(feature = $feature)]
         #[proc_macro_derive($trait_, attributes($($attribute),*))]
-        #[doc(hidden)]
+        #[doc = include_str!(concat!("../doc/", $feature, ".md"))]
```
JelteF, line 118: I think this is as alright solution for now, but I'm not a fan of the duplication this causes. The `Add` doc is essentially copy pasted for every `Add`-like derive:
image

I'm not sure how to make this better without making it worse in other ways. So I'm fine with this for now.

src/lib.rs
```diff
@@ -1,185 +1,30 @@
...
+//! [`IndexMut`]: crate::IndexMut
+//! [`DerefMut`]: crate::DerefMut
+//! [`AddAssign`-like]: crate::AddAssign
+//! [`MulAssign`-like]: crate::MulAssign
+//!
+//! [`Constructor`]: crate::Constructor
+//! [`IsVariant`]: crate::IsVariant
+//! [`Unwrap`]: crate::Unwrap
+#![doc = include_str!("../README.md")]
+#![cfg_attr(docsrs, feature(doc_auto_cfg))]
```
JelteF, line 27: What is this suppossed to do?

JelteF (changes requested):
Thank you for working on this! I think it's almost perfect.

src/lib.rs
JelteF: Okay after a bit of research and playing around, I realised it's to shows the `feature` that enables a derive. Is this the only thing, or am I missing something. (showing the feature does seem very useful btw).

README.md
tyranron: @JelteF yes, I've tried but haven't found a way around it. Still, I think having this line exposed in GitHub's README is a lesser evil than duplication and copy-squating.

README.md
tyranron: @JelteF nope. 

1. These links do work in `README.md` only and are overwritten on the crate level for the generated Rust docs. You can run `cargo doc --open` and see that they point correctly there.

2. As I've pointed out in this PR description, the `doc/*.md` solution is temporary, to not break the README on current `master` after merging this PR, as there are no such docs at `docs.rs` at the moment. Once we release `1.0.0` they can be changed to point to `docs.rs` directly.

src/lib.rs
tyranron: @JelteF yup, this is solely for the "Available on crate feature `<feature_name>` only" badges on `docs.rs`.

README.md
JelteF: My main problem is that we want relative path links on docs.rs to make sure the docs main page for old derive_more docs keeps pointing to sub pages for that same version. But the readme should instead use full urls. I don't see any way around this. I also don't like copy pasting docs, but I think it's the least bad option. Another advantage of it is that the readme can stay valid for the previous version of derive_more during development of the next version.

tyranron: The only issue is with the [hidden line exposed on GitHub in example code for 2015 edition](https://github.com/JelteF/derive_more/pull/209#discussion_r996359502).

tyranron: @JelteF ping

JelteF: > After we rollout 1.0.0 release, the links in README on GitHub will point to https://docs.rs/derive_more/latest which is exactly what we need. On docs.rs itself they will remain relative.

Maybe I'm missing something, but that sounds like for every future release we'll have this problem. And we need to do the dance with swapping the links in the readme before we release and after we release.

I think a better way is to have another file than README.md being the source of lib.rs docs. And have a simple `sed` command in a script to transform this to our README. Thus making the README a build artifact of the cannonical docs file.

Something like this:
```bash
sed -e 's=: doc/=: https://docs.rs/derive_more/latest/=g' -e '/# fn main() {}/d' doc/index.md > README.md
```

JelteF: One thing I dislike about the current PR is that it breaks the README until we release.

tyranron: @JelteF as I've mentioned before, the only issue is the `# fn main() {}` line.

> One thing I dislike about the current PR is that it breaks the README until we release.

Why it breaks? It leads to actual `doc/*.md` files. The links work as expected.

If you want them to point to the previous location (`https://jeltef.github.io/derive_more/derive_more/*.html`), then it's OK too. I'll change them until we do the release.

> Maybe I'm missing something, but that sounds like for every future release we'll have this problem.

No. Once new docs hit the `docs.rs`, we change the links in README to `https://docs.rs/derive_more/latest/`, and for any subsequent releases we don't need to do anything.

> And we need to do the dance with swapping the links in the readme before we release and after we release.

No, we don't need to swap anything before a release. We only need to swap links in README after the first `1.0.0` release once, and only because external docs don't exist on `docs.rs` at the moment.

> I think a better way is to have another file than README.md being the source of lib.rs docs. And have a simple bash script with a `sed` command to transform this to our README. Thus making the README a build artifact of the cannonical docs file.

Yes, I've thought about it. But as I've [mentioned before](https://github.com/JelteF/derive_more/pull/209#discussion_r996430435), I think that a single `# fn main() {}` line is not worth this tooling complication.

tyranron: > If you want them to point to the previous location (`https://jeltef.github.io/derive_more/derive_more/*.html`), then it's OK too. I'll change them until we do the release.

Done in https://github.com/JelteF/derive_more/pull/209/commits/c1a11fd5be78f84384b21f48eec3ff26db935957.

src/lib.rs
```diff
@@ -1,185 +1,30 @@
...
+//! [`Mul`-like]: crate::Mul
+//! [`Sum`-like]: crate::Sum
+//! [`IndexMut`]: crate::IndexMut
+//! [`DerefMut`]: crate::DerefMut
+//! [`AddAssign`-like]: crate::AddAssign
+//! [`MulAssign`-like]: crate::MulAssign
+//!
+//! [`Constructor`]: crate::Constructor
+//! [`IsVariant`]: crate::IsVariant
+//! [`Unwrap`]: crate::Unwrap
```
tyranron, line 25: @JelteF please, pay the attention to these links above. They do overwrite the ones in README, to work properly as relative intra-doc links in the generated Rust docs (either on `docs.rs` or via `cargo doc`).

tyranron: I guess I should add a comment explaining why they're there.

tyranron: Done in https://github.com/JelteF/derive_more/pull/209/commits/545f737449562cd5c94aa32a30ca95be73d3ee98.

JelteF: Cool that's a smart trick! I indeed missed that. Thanks for adding the explain comment.

README.md
JelteF: Thanks for explaining and reverting the links, and yeah I agree now this is a good solution. I didn't realise your link overwriting trick.

For clarification though, I don't think the doc/*.md links would have gone to the github files.

They should have been blob/master/doc/*.md links instead:
1. https://github.com/JelteF/derive_more/blob/master/doc/add.md
2. https://github.com/JelteF/derive_more/doc/add.md

JelteF (approved):

PR #213, opened 2022-10-20
Include docs in package

JelteF:
When testing out `cargo release patch` (dry-run by default) I would get
the following errors:

```
error: couldn't read src/../doc/mul_assign.md: No such file or directory (os error 2)
   --> src/lib.rs:120:17
    |
120 |           #[doc = include_str!(concat!("../doc/", $feature, ".md"))]
    |                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
...
191 | / create_derive!(
192 | |     "mul_assign",
193 | |     mul_assign_like,
194 | |     ShrAssign,
195 | |     shr_assign_derive,
196 | |     shr_assign,
197 | | );
    | |_- in this macro invocation
    |
    = note: this error originates in the macro `include_str` which comes from the expansion of the macro `create_derive` (in Nightly builds, run with -Z macro-backtrace for more info)
```

This is because the doc files are now included in the sources, but they
were not part of the package. This fixes that.


tyranron (approved):

emilio:
That means that folks vendoring the crate now need to include all docs and tests which seems unfortunate :(

Issue #131, opened 2020-06-14
Constructor generated new should be marked const


KarelPeeters:
For example, given this struct:
```rust
#[derive(Constructor)]
struct Foo { x: f32, y: f32 }
```
The generated code should be:
```rust
impl Foo {
    const fn new(x: f32, y: f32) -> Self {
        Self { x, y }
    }
}
```

As far as I know this should be fine to do on the current stable. It's an easy change, would a PR adding this be accepted?

JelteF:
A pull request would be accepted for this assuming it doesn't break Rust 1.36.0 (which is the currently lowest supported Rust). To do so you might have to check the version with `rustc_version` in the `build.rs` script.

flaghacker:
Unfortunately `const fn` is not quite as stable as I thought, for example trait bounds aren't yet supported in stable (https://github.com/rust-lang/rfcs/pull/2632). I think this issue will have to wait until `const fn` is more mature.

krojew:
Any plans for this? Having const constructors would be a nice feature.

appetrosyan:
One option for how it could be rendered is if we add `#[CosntConstructor]` to the derive macros. Then if the `const`-ness is not supported you have alternatives. No need for complex logic, just a conscious choice.

Issue #202, opened 2022-10-04
Wait for Rust 1.65 for our 1.0 release


JelteF:
The 1.0 release is pretty much ready from our side. So I was planning to release it in the next few days. But after reading up on the `let_else` Rust feature, I learned that this will be stabilized in Rust 1.65, which is early November.

Personally I think this feature can make the code base much more readable. So, I'm now planning to wait for the 1.65 release and set our MSRV to 1.65 for derive_more 1.0.

@tyranron does that sound good to you?

phip1611:
Hey! In https://github.com/JelteF/derive_more/pull/192#issuecomment-1152091758 @tyranron said that you should add `#[automatically_derived]` to several places. Seems like this didn't happen yet. Any comment?

tyranron:
@JelteF yes. Sounds, good to me.

@phip1611 yup, there was an issue disallowing merging PRs properly due to CI. Now it's resolved, so we'll cover the things shortly.

> PS: I just found this. https://github.com/JelteF/derive_more/blob/master/Cargo.toml#L71 you might want to change that (at least the comment) as well, as the MSRV is bumped?

Thanks, will take a look into.

tyranron:
@phip1611 

> Hey! In [#192 (comment)](https://github.com/JelteF/derive_more/pull/192#issuecomment-1152091758) @tyranron said that you should add `#[automatically_derived]` to several places. Seems like this didn't happen yet. Any comment?

Done in https://github.com/JelteF/derive_more/pull/203.

tyranron:
@phip1611 

> PS: I just found this. https://github.com/JelteF/derive_more/blob/master/Cargo.toml#L71 you might want to change that (at least the comment) as well, as the MSRV is bumped?

Done in https://github.com/JelteF/derive_more/pull/204.

tyranron:
@JelteF before going 1.0 I would like to propose more breaking changes.

## 1. Rename `CustomDebug` as `DebugCustom` (or even `DebugSmart`)

[...]

PR #218, opened 2022-11-10
Add `const` to constructors by a feature

fmiguelgarcia:
Resolves [Constructor generated new should be marked const](https://github.com/JelteF/derive_more/issues/131)

## Synopsis

Added a new `const` feature which generates `const fn` on `Constructor` derives.

## Checklist

- [x] Tests are added/updated **IMPORTANT**: I am not sure if new tests are added to the CI.
- [x] [CHANGELOG entry][l:1] is added


tyranron:
Should this really be feature-gated? I cannot imagine the situation where `const` will be harmful on constructors.

fmiguelgarcia:
> Should this really be feature-gated? I cannot imagine the situation where `const` will be harmful on constructors.

Just in case you are using a legacy compiler with no support of `const` in Generics.
Having said that, I think `const` feature should be enable by default, WDYT?

att. @tyranron

tyranron:
@fmiguelgarcia well, once we release 1.0 (next release) we will have [MSRV set to 1.65](https://github.com/JelteF/derive_more/issues/202), so you won't be able to use legacy compilers 🤷‍♂️

JelteF:
Also, I feel like the same should be done for the `IsVariant` and `Unwrap` derives.

tyranron:
@JelteF `IsVariant` yes, but `Unwrap` definitely not, as `const` at the moment is not very clever with destructor evaluation checks, so almost any struct/enum destruction in `const` leads to compilation error. See https://github.com/rust-lang/rust/issues/60964 for details.

fmiguelgarcia:
@tyranron I've used the `const` feature to enable const in `Unwrap` (I've just read your comment).
`IsVariant` and `Constructor` now uses `const fn`.

Do you want to remove `const` feature completely? Or Do you want to keep it and support `const` in `Unwrap`? 
att. @JelteF @tyranron

tyranron:
@fmiguelgarcia I'd vote for removing `const` feature completely. We always may `const`ify `Unwrap` once `feature(const_precise_live_drops)` is stabilized. It's not something that vital, deserving to squat-around it with an additional Cargo feature.

JelteF:
I think this is almost good to merge, just a few things are needed:
1. Remove const from unwrap due to the bugs (so only const for constructor and isvariant for now)
2. Remove the const feature and simply always use const.
3. fix clippy errors.

JelteF:
@fmiguelgarcia do you have time to do that? This is one of the final things left on the 1.0.0 milestone.

JelteF (approved):

PR #220, opened 2022-11-14
Use dedicated error type for TryInto errors

### LINKED ISSUE #173: Make TryInto use better error type
Rather than returning a string as an error, might it not be better to use a struct that stores the type and variant names as strings, which implements Display to show a full error message? That struct can then easily implement `std::error::Error` too.

Perhaps this can be enabled as an optional feature, as suggested in https://github.com/JelteF/derive_more/pull/130? Although it kinda feels like this should be the default behaviour

JelteF:
Instead of returning a string as the error, this returns a dedicated
error type. This has the big benefit that in case of an error it's
possible to get back the original value by using `error.input`.
Previously the original value would be dropped in case of an error.

Fixes #173
Fixes #130

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,6 +88,8 @@ jobs:
              | sh -s -- --git hobofan/cargo-nono
                         --tag $(curl -s https://api.github.com/repos/hobofan/cargo-nono/releases/latest
                                 | jq -r '.tag_name')
+             # TODO: Remove the latest Git tag detection above once this PR is merged:
+             #       https://github.com/japaric/trust/pull/137
 
       - run: cargo nono check --package derive_more
                    --no-default-features --features all_no_std
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -86,6 +86,8 @@ jobs:
       - name: Install cargo nono
         run: curl -LSfs https://japaric.github.io/trust/install.sh
              | sh -s -- --git hobofan/cargo-nono
+                        --tag $(curl -s https://api.github.com/repos/hobofan/cargo-nono/releases/latest
+                                | jq -r '.tag_name')
 
       - run: cargo nono check --package derive_more
                    --no-default-features --features all_no_std
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -75,6 +75,21 @@ jobs:
 
       - run: cargo test --workspace --features testing-helpers
 
+  no_std:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: dtolnay/rust-toolchain@v1
+        with:
+          toolchain: stable
+
+      - name: Install cargo nono
+        run: curl -LSfs https://japaric.github.io/trust/install.sh
+             | sh -s -- --git hobofan/cargo-nono
+
+      - run: cargo nono check --package derive_more
+                   --no-default-features --features all_no_std
+
   test:
     strategy:
       fail-fast: false
@@ -146,6 +161,7 @@ jobs:
     needs:
       - clippy
       - msrv
+      - no_std
       - rustdoc
       - rustfmt
       - test
--- a/src/errors.rs
+++ b/src/errors.rs
@@ -2,13 +2,13 @@ use core::fmt;
 
 /// Error returned by the derived [`TryInto`] implementation.
 ///
-/// [`TryInto`]: crate::TryInto@macro
+/// [`TryInto`]: macro@crate::TryInto
 #[derive(Clone, Copy, Debug)]
 pub struct TryIntoError<T> {
     /// Original input value which failed to convert via the derived
     /// [`TryInto`] implementation.
     ///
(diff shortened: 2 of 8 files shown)


src/errors.rs
```diff
@@ -0,0 +1,32 @@
+use core::fmt;
+
+#[derive(Debug)]
```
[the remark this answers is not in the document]
tyranron: @JelteF 

> Open question 1: Do we to derive `Debug`?

Yes. We do want to derive `Debug` (and `Clone` too), because it's so unpleasant situation when third-party lib doesn't provide a `Debug` impl for a type, and you should squat-around it with wrappers in case you just want to put `dbg!()`, and I hit it so many times 🤬

> Maybe not, because it requires all types for which TryFrom is derived to also have a Debug implementation. Can we make this conditionally derive Debug only when T implements Debug itself?

Yup. Event standard `#[derive(Debug)]` will generate a proper `impl<T: Debug> Debug for MyError<T>` in this situation. We don't need to do something additional here.

> Open question 2: Do we want to implement the Error trait for this error type? This doesn't work for no_std currently, but we could make std an optional default enabled feature.

Yes. Definitely.

src/errors.rs
```diff
@@ -0,0 +1,32 @@
...
+pub struct TryIntoError<T> {
```
tyranron, line 4: I'd like to add docs for this error type as it's part of public API now.

src/lib.rs
```diff
@@ -32,3 +32,6 @@
 
 #[doc(inline)]
 pub use derive_more_impl::*;
+
+mod errors;
+pub use crate::errors::TryIntoError;
```
tyranron, line 37: Also, I think this should be feature-gated by `"try_into"`.

tyranron (changes requested):

src/errors.rs
```diff
@@ -0,0 +1,32 @@
...
+    pub input: T,
+    variant_names: &'static str,
+    output_type: &'static str,
+}
+
+impl<T> TryIntoError<T> {
+    pub fn new(
```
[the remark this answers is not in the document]
tyranron: Done in https://github.com/JelteF/derive_more/pull/220/commits/55a4268a82cd6aef704429729a85192e9f0e3aec.

src/errors.rs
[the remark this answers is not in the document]
tyranron: Added deriving `Clone` and `Copy` in https://github.com/JelteF/derive_more/pull/220/commits/a6e68dfb07552ba881d658cd2d23991b1eaf5ce5. These are conditional in the same way as `#[derive(Debug)]`.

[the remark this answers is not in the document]
JelteF: > Yup. Event standard #[derive(Debug)] will generate a proper impl<T: Debug> Debug for MyError<T> in this situation. We don't need to do something additional here.

Yes of course, it seems I was misreading the following error message when removing `Debug` from `MixedData`:
```
error[E0277]: `MixedData` doesn't implement `Debug`
    --> src/lib.rs:250:42
     |
14   | assert_eq!(123u32, mixed_int1.try_into().unwrap());
     |                                          ^^^^^^ `MixedData` cannot be formatted using `{:?}`
     |
     = help: the trait `Debug` is not implemented for `MixedData`
     = note: add `#[derive(Debug)]` to `MixedData` or manually `impl Debug for MixedData`
     = help: the trait `Debug` is implemented for `TryIntoError<T>`
     = note: required because of the requirements on the impl of `Debug` for `TryIntoError<MixedData>`
 ```

the reason it's complaining here is because `unwrap` requires an `E` that implements `Debug`

File: src/errors.rs (hunk 2 above)
Comment (tyranron) [reply src/errors.rs:7]: Done in https://github.com/JelteF/derive_more/pull/220/commits/ae3d5c05cee2a8c564edac4d2edc38239319f5f9.

File: src/lib.rs (hunk 3 above)
Comment (tyranron) [reply src/lib.rs:39]: Done in https://github.com/JelteF/derive_more/pull/220/commits/77043d1fd53825cb09eb3ada86f1e1b11853140f.

File: .github/workflows/ci.yml [hunk 5]
```diff
@@ -75,6 +75,25 @@ jobs:
 
       - run: cargo test --workspace --features testing-helpers
 
+  no_std:
```
tyranron, line 78: @JelteF I've added this CI check, so we can track possible `no_std` regressions on CI.

tyranron (approved):
🎉

.github/workflows/ci.yml
JelteF: I'm supprised it works, because this issue looks like it would impact this crate: https://github.com/hobofan/cargo-nono/issues/47

tyranron: @JelteF the reason is [explained](https://github.com/hobofan/cargo-nono/issues/47#issuecomment-1197767742):
> there is just no logic in place yet to track and evaluate compiler switches on individual use items (as evaluating arbitrary compiler switches was a rabbit hole I didn't want to go into).

We just have no compiler switches on `use` statements at the moment. So, we're lucky.

PR #222, opened 2022-11-14
Update MSRV to 1.65

### LINKED ISSUE #202: Wait for Rust 1.65 for our 1.0 release

JelteF:
Rust 1.65 adds support for `let ... else`. This is quite a useful
feature for this library since early exits often make a lot of sense
when matching against AST enums.

Fixes #202


tyranron (approved):
@JelteF please, correct the required CI checks for `master` branch.

Anything else seems fine.

Issue #216, opened 2022-11-09
Serde like `rename` for Display & FromStr trait on enum


ModProg:
i.e. I would expect
```rs
#[derive(Display)]
#[display(rename="snake_case")]
enum Test{
   First
}

println!("{}", Test::First) // -> prints "first"
```

The serde behaviour is documented here:
https://serde.rs/container-attrs.html#rename_all

> ### #[serde(rename_all = "...")]
> Rename all [...] variants [...] according to the given case convention. The possible values are "lowercase", "UPPERCASE", "PascalCase", "camelCase", "snake_case", "SCREAMING_SNAKE_CASE", "kebab-case", "SCREAMING-KEBAB-CASE".

JelteF:
This sounds useful. It fits well with the FromStr for unit enum support that was introduced in #176. 

bikeshed: I think I prefer `rename` over `rename_all`. That way you could also put `rename` on a single field. The top level `rename` would then simply set the default for enum items.

JelteF:
A good question raised in the thread in #225 would be what should be the default? My summary of that discussion is that there are two different usages for which a different default makes sense:
1. Error types: you probably want a custom error message. By requiring to set this attribute you would not be able to forget to set a custom message for one of your variants.
2. Any other enum type: You probably want nice display by default

I'd say go for strict by default for now, i.e. error by default. Then if that is deemed too cumbersome we can always relax the default. However, unrelaxing the default would be a breaking change.

ilslv:
@JelteF 

> bikeshed: I think I prefer rename over rename_all. That way you could also put rename on a single field. The top level rename would then simply set the default for enum items.

Actually `#[display(rename("..."))` is an alias for direct `#[display("...")]`, so I would vote for `rename_all` as seen in `serde` and already widespread across the ecosystem. 

> I'd say go for strict by default for now, i.e. error by default. Then if that is deemed too cumbersome we can always relax the default. However, unrelaxing the default would be a breaking change.

I think that going explicitly over implicitly is a viable option, but only when it's consistent. We already have 2 examples for implicit display:

```rust
#[derive(Display)]
struct Unit;

#[derive(Display)]
struct SingleField(u8); // works for named `{ field: u8 }` too
```

I see implicit `Display` impl on enums with unit fields as understandable and discoverable continuation for that:

```rust
#[derive(Display)]
enum Enum {
    A,
    B,
}
```

[...]

PR #225, opened 2022-11-18
Redesign `Display` derive macro

### LINKED ISSUE #142: `affix` is confusing with outer-level `enum` display formatting
Currently, when one uses a top-level `display` attribute for the `Display` derive on an `enum`, it switches to an affix mode that defines the `enum`'s overall format and expects, at most, a single placeholder where inner variant display logic can print into:

```rust
#[derive(Display)]
#[display(fmt = "Prefix {} Suffix")] // at most 1 placeholder allowed here
enum Enum {
    #[display(fmt ="value")] // prints "Prefix value Suffix"
    Variant1,
    // ...
}
```

However, there's a few problems with this current behavior:

* It's REALLY surprising to discover that `fmt` works rather differently than anywhere else here. It's not obvious why formatting _should_ be different here, and there's no "obvious" signal that behavior is different. I feel qualified to say this since _I implemented this behavior myself a year ago_, and still was confused after reading diagnostics until I remembered what I had implemented.
* Affix formatting defined this way is currently all-or-nothing. Once the design of the `Display` impl requires using a specific format some of the time,

I think that we can do better here. There are enough surprises with the current behavior that I think exploring alternatives is warranted. Here's what we stand to gain:

* We can develop an API that operates as similarly to variant-level formatting as possible, which makes the design more coherent.
* Because `derive_more` inhabits the same ecosystem as other crates that do `Display` deriving with rather different approaches, we could possibly try to find a top-level `enum` formatting option that is also consistent with those approaches (and play a bit better with other crates conceptually). In a second, I'll propose something similar to what `thiserror`, an error crate with facilities for deriving `Display`, currently does.

So, here's my proposal:

* The current affix-oriented behavior should be moved to an `affix` attribute key, and (this) new behavior will be defined for `fmt` instead.
* Re-define the `fmt` spec at the top level of an `enum` to be the _default_ printing logic for an `enum`'s variants that can be overridden by a more specific `fmt` rule. An overridable default has basis in some prior art; `thiserror::Error` treats top-level `enum` format specifiers as the "default" message ([motivating issue](https://github.com/dtolnay/thiserror/issues/18), [current relevant upstream `master` code at time of writing](https://github.com/dtolnay/thiserror/blob/04b91d76462834279a323b7771b48158aba19db4/impl/src/ast.rs#L83)):

    ```rust
    use thiserror::Error;
    #[derive(Debug, Error)]
    #[error("error")]
    enum Enum {
        A, // displays "error"
        #[error("b")]
        B, // displays "b"
    }
    ```
* Disallow `affix` and `fmt` to be used together at the same time for the time being, since that's a lot of open design space that doesn't need to be handled right this second.

To concretely translate my above suggestions into code:

```rust
use derive_more::Display;

// Using affixes at the top level of an `enum` is the same as before, except we
// use an `affix` key in the `display` attribute instead of `fmt`.
#[derive(Display)]
#[display(affix = "<{}>")]
enum UsesAffix {
    #[display(fmt = "a")]
    A, // prints "<a>"
    #[display(fmt = "b")]
    B, // prints "<b>"

	// This makes no sense and is still disallowed.
	// #[display(affix = ...)]
	// C,

    // ...
}

// A top-level `display(fmt = ..., ...)` now is:
//
// * Just like variant-level `display` attributes in that it uses formatting
//     args as usual. Only `self` is bound in arguments' context, of course.
// * Overridable by a variant-level `display` attribute.
//
// One motivating example for this is to just use `Debug` printing
// by default unless we happen to have something better.
#[derive(Display)]
#[format(fmt = "{:?}", self)]
enum DisplayDefaultsToDebug {
	A, // prints "A"
	#[display(fmt = "Display B")]
	B, // prints "Display B"
}
```

---

I'm working on a draft PR for this [here](https://github.com/JelteF/derive_more/pull/143). I'm speculatively doing work without getting validation in this issue first, but I don't mind. :)

### LINKED ISSUE #217: Consider using non string expressions
This woud allow easier writing of expressions like:
```rs
#[display(fmt = "Focus {_0} {}", "if *_1 {\"global\"} else {\"\"}")]
struct Fullscreen(Edt, bool);
```
which becomes, also improving highlighting for types such as strings
```rs
#[display(fmt = "Focus {_0} {}", if *_1 {"global"} else {""})]
struct Fullscreen(Edt, bool);
```
Clap switched to this with their 4.0 release: https://docs.rs/clap/latest/clap/_derive/index.html

There is also a helper crate that could make parsing these easier: https://lib.rs/crates/attribute-derive (Disclaimer, I have written this crate).

ilslv:
Resolves #217 

## Solution

1. Use expressions instead of string literals for fmt arguments: 
   `#[display("{}", "crab.rave()")]` -> `#[display("{}", crab.rave())]`

## Checklist

[l:1]: /CHANGELOG.md


ilslv:
@tyranron @JelteF so we are on the same page, let me go over design decisions, that are already set in stone, before final decision on #142 is made.

This should cover ***every*** possible use-case, everything else should result in compile-time error. For example:

```rust
#[derive(Display)]
#[display("Field: {}")] // error: 1 positional argument in format string, but no arguments were given
struct Single(i32);
```

# Structs

## Unit

```rust
#[derive(Display)]
struct Unit; // "Unit"

#[derive(Display)]
struct Tuple (); // "UnitTuple"

#[derive(Display)]
struct Struct {} // "UnitStruct"
```

## Single-field

```rust
#[derive(Display)]
struct Tuple(i32); // "1"

#[derive(Display)]
struct Struct { field: i32 } // "1"

#[derive(Display)]
#[display("Pre: {_0}")]
struct PreTuple(i32); // "Pre: 1"

#[derive(Display)]
#[display("Pre: {field}")]
struct PreStruct { field: i32 } // "Pre: 1"
```

## Multi-field

```rust
#[derive(Display)]
#[display("{_0} {ident} {_1} {} {}", _1, any_expr(_0, _1), ident = 123, _1 = _0)]
struct Tuple(i32, i32) // "1 123 1 2 3"

#[derive(Display)]
#[display("{field1} {ident} {field2} {} {}", field2, any_expr(_0, _1), ident = 123, field2 = field1)]
struct Struct { field1: i32, field2: i32 } // "1 123 1 2 3"
```

# Enums

## Without variants (Infallible-like)

```rust
#[derive(Display)]
enum Void {}

const fn has<T: Display>() {}
const _ = has::<Void>();
```

## Unit variant

```rust
#[derive(Display)]
enum Enum {
    First, // "First"
    FirstTuple(), // "FirstTuple"
    FirstStruct {}, // "FirstStruct"
    #[display("SECOND")]
    Second, // "SECOND"
    #[display("SECOND_TUPLE")]
    SecondTuple(), // "SECOND_TUPLE"
    #[display("SECOND_STRUCT")]
    SecondStruct {}, // "SECOND_STRUCT"
}
```

## Single-field variant

```rust
#[derive(Display)]
enum Enum {
    First(i32), // "1"
    FirstNamed  { field: i32 }, // "1"
    #[display("Second: {_0}")]
    Second(i32), // "Second: 1"
    #[display("Second: {field}")]
    SecondNamed { field: i32 }, // "Second: 1"
}
```

## Multi-field variant

```rust
#[derive(Display)]
enum Enum {
    #[display("{_0} {ident} {_1} {} {}", _1, any_expr(_0, _1), ident = 123, _1 = _0)]
    Tuple(i32, i32), // "1 123 1 2 3"
    #[display("{field1} {ident} {field2} {} {}", field2, any_expr(_0, _1), ident = 123, field2 = field1)]
    Struct { field1: i32, field2: i32 }, // "1 123 1 2 3"
}
```

## Interactions with top-level attribute

TODO: Fill out, once #142 is resolved

# Unions

```rust
#[derive(Display)]
#[display("No field accesses are allowed, but {} is ok", "interpolation")]
union Union { // No field accesses are allowed, but interpolation is ok
    field: i32,
}
```

# Unresolved questions

Should we support `{0}` syntax in addition/instead of `{_0}`. `thiserror` does that.
Personally, I'm in favour of this feature in addition to legacy syntax. I think we can even emit deprecated warnings on them.

ilslv:
Discussed with @tyranron: Merge this PR without support for top-level enum attribute and do it as a separate PR.

tyranron:
@ilslv @JelteF one thing I still dislike is semantics for unit types:
```rust
#[derive(Display)]
struct Unit; // "Unit"

#[derive(Display)]
struct Tuple (); // "UnitTuple"

#[derive(Display)]
struct Struct {} // "UnitStruct"

#[derive(Display)]
enum Enum {
    First, // "First"
    FirstTuple(), // "FirstTuple"
    FirstStruct {}, // "FirstStruct"
}
```

This is too subtle and implicit. And may be too unexpected if I write `derive(Octal)` or similar.

I think it's better to return error in such cases, pushing a user to explicitly specify what he wants. Giving a convenient top-level attribute semantics, there would be almost no ergonomics impact if needs exactly this behaviour:
```rust
#[derive(Debug, Display)]
#[display("{self:?}")]
struct Unit; // "Unit"

#[derive(Debug, Display)]
#[display("{self:?}")]
struct Tuple (); // "Tuple"

#[derive(Debug, Display)]
#[display("{self:?}")]
struct Struct {} // "Struct"

#[derive(Debug, Display)]
#[display("{self:?}")]
enum Enum {
    First, // "Enum::First"
    FirstTuple(), // "Enum::FirstTuple"
    FirstStruct {}, // "Enum::FirstStruct"
}
```

But I would argue that most of the time, a user doesn't expect something `Debug`-like on `Display`ing unit types.

ilslv:
@tyranron 

> This is too subtle and implicit. And may be too unexpected if I write derive(Octal) or similar.

Valid point, if something like this exists, it should be only for `Display`.

---

I want to explain, why I've added this behaviour in a first place. 
From my personal experience, it's very unpleasant to write something like this by hand:

```rust
// #[derive(Display)]
enum Enum {
    A, 
    B,
    C,
}

impl fmt::Display for Enum {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        match self {
            Self::A => "A".fmt(f),
            Self::B => "B".fmt(f),
            Self::C => "B".fmt(f),
        }
    }
}
```

This is a lot of tedious code to write and copy-pasting may lead to errors. Have you noticed, that code block above contains an error?

I consider this feature desired, understandable and easily **discoverable** just by writing code without looking into the docs. Emphasis on **discoverable**, because this is one of the most delightful features of Rust to me. Discoverable intuitive design with compile-time checks is a match made in heaven. I don't think you can misinterpret it and your suggested approach doesn't cover this use-case. 
For unit structs this feature is less desirable, but I would add it just for symmetry. `#[display("{self:?}")]` looks really scary and something like only maintainer with a deep knowledge of this crate will use 🙃

tyranron:
@ilslv 

> `#[display("{self:?}")]` looks really scary and something like only maintainer with a deep knowledge of this crate will use 🙃

Isn't this the first thing a rustacean would try (or write manually) when he thinks about "`Display` as `Debug`" semantics?

> I don't think you can misinterpret it 

I do think such a feature implicitly pushes a library user to mix `Display`/`Debug` semantics, which is usually not a good idea. Given that most use-cases of this macro, in practice, is defining custom error types, I would prefer a compilation error about forbidden error description, rather than implicitly making it `Debug`-like.

> and your suggested approach doesn't cover this use-case.

Ah, you're talking more about [`strum::Display` semantics](https://docs.rs/strum/0.24/strum/derive.Display.html), rather than "`Display` as `Debug`" semantics, so we can replace `strum::Display` with `derive_more::Display` naturally. Now, what you're talking about makes much more sense. 🤔

Yeah, restricting it to `Display` only, has sense. But I would like to hear other opinions as well.  
ack @JelteF @ModProg

ilslv:
@tyranron 

> Isn't this the first thing a rustacean would try (or write manually) when he thinks about "Display as Debug" semantics?

To be fair I didn't think about this possibility, when implementing this functionality. And even if I just wasn't thinking enough, this assumption applies only to experienced rustaceans, but this library isn't only for them with ~50k daily downloads. 

> Given that most use-cases of this macro, in practice, is defining custom error types

I wouldn't just jump to this assumption, especially when there is `thiserror`. General-purpose `Display`s should be at the same level of priority. 

> so we can replace strum::Display with derive_more::Display naturally

Exactly!

tyranron:
@ilslv 

> Exactly!

And then, we [want to play with casing](https://github.com/JelteF/derive_more/issues/216). Maybe we could opt-in this behaviour explicitly by specifying `#[display(rename_all = ...)]`?

ilslv:
@tyranron i meant out of "scope for this PR" 🤦‍♂️
But definitely would like to see that functionality.

tyranron:
@ilslv I'm not talking about customizing, but rather still giving options how we can deal with `Debug`-like defaults for units, which is in scope of this PR.

ModProg:
I'm not sure what the right default behavior is for derive Display. 

> @ilslv I'm not talking about customizing, but rather still giving options how we can deal with `Debug`-like defaults for units, which is in scope of this PR.
> 
> ```rust
> #[derive(Display)]
> #[display(rename_all = "upper")]
> enum Enum {
>     A,  // "A"
>     B,  // "B"
>     C,  // "C"
> }
> 
> #[derive(Display)]
> enum AnotherEnum {  // compilation error!
>     A, 
>     B, 
>     C, 
> }
> ```

The advantage of this behavior is, it removes any ambiguity on what the output is. Though, it would in my opinion make sense to adopt serde's casing for the `rename_all` to make it 100% clear what is happening: 

We could also add a way to add something like `rename_all = "Debug"` but that is not necessarily better than just saying `{self:?}`

One thing I just thought about would be to somehow enable the renamed variant in the `default_format` of an enum, kind of like so:
```rs
#[display("{self:snake_case} {_0}")]
```
But that is probably a very niche feature, and not really worth increasing complexity for (only thought about it, because I'd have a use in a current project).

ModProg:
> To be fair I didn't think about this possibility, when implementing this functionality. And even if I just wasn't thinking enough, this assumption applies only to experienced rustaceans, but this library isn't only for them with ~50k daily downloads.

This should definitely be documented in the Display derive macro documentation, as I feel like this might be a generally quite useful usage.

ilslv:
@tyranron 

> ```rust
> #[derive(Display)]
> enum AnotherEnum {  // compilation error!
> ```

I vote against explicitly specifying `#[display(rename_all = "upper")]`. I don't think this gives us some value, only adds friction.

JelteF:
To be clear, I haven't yet looked at the code. I think the Unit enum stuff sounds useful, but shouldn't be part of this PR. Also introducing it doesn't break the Display derive API, so I don't think it should be part of the 1.0.0 milestone. I think we should move discussion about that feature to #216

ModProg:
@tyranron generally sound good, only thing it cannot unite is my "I want an error when I forget to add a format to an enum Variant".

But that would've been a breaking change anyway.

tyranron:
@ModProg 

> only thing it cannot unite is my "I want an error when I forget to add a format to an enum Variant".

Why? Isn't this case about this?
```rust
#[derive(Error)] // look, ma! we don't need to specify obvious `Display` here
enum AnotherError {
    A,          // expansion error, as we forgot to specify error description here
    B(String),  // prints the inner string
}
```

Or you've meant any enum variant, not just a unit-like?

ModProg:
@tyranron yes, and my usecase wasn't an error. It was generating commands.

ModProg:
> Moreover I don't really think "forgetting to specify formatting" is a big deal. In this case error message may confuse you for a second, but code itself should help pin down the cause for an error.

Well if it is an error that is, in different usecases this might be a bigger issue.


But I do understand that for most users this is more convenient than requiring them always.

Maybe it could be a default-feature that can be disabled? But in the end, I'm fine with anything.

JelteF:
@tyranron While I agree that Error deriving display is more ergonomic, it's not in line with how derives normally work afaik. At least in the stdlib it's not how it works. `Eq` alse requires deriving `PartialEq`. Also, I think such a proposed subtle difference between how `derive(Error)` derives `Display` and how `derive(Display)` would cause confusion itself. Especially since (in your example at least), one uses `error(...)` while the other would use `display(...)` as the attribute.

After all this extra information and discussion, I'm now siding with @ilslv and I agree we should derive units by default.

JelteF:
> Maybe it could be a default-feature that can be disabled? But in the end, I'm fine with anything.

That seems like a non-breaking change (since it's opt-in), so I think we can postpone that decision until after 1.0

tyranron:
@JelteF thank you for the feedback. You and @ilslv convinced me that the current behaviour would be a lesser evil.

I'll review this PR shortly, so we can move towards  releasing 1.0.

tyranron (dismissed):
> I'll review this PR shortly.

> shortly

_shortly_™... 😅

@ilslv what an amazing work. Thanks!

I've adjusted docs a little and some local variables naming. Everything else seems good.

tyranron:
@JelteF would you like to take a look on this? Or we can merge it?

The CI issue is fixed in https://github.com/JelteF/derive_more/pull/231.

JelteF:
I merged #231 if this is rebased I guess CI should pass now. Feel free to merge it.

impl/doc/display.md
```diff
@@ -77,25 +70,22 @@ The following where clauses would be generated:
 ### Custom trait bounds
 
 Sometimes you may want to specify additional trait bounds on your generic type parameters, so that they
-could be used during formatting. This can be done with a `#[display(bound = "...")]` attribute.
+could be used during formatting. This can be done with a `#[display(bound(...))]` attribute.
 
-`#[display(bound = "...")]` accepts a single string argument in a format similar to the format
+`#[display(bound(...))]` accepts a single string argument in a format similar to the format
```
ModProg, line 75: I don't think it is a string anymore is it?

ModProg (changes requested):

tyranron (approved):

PR #226, opened 2022-11-18
Tune up code style

tyranron:
Related to #202, #222

## Synopsis

Since we've [bumped MSRV to 1.65](https://github.com/JelteF/derive_more/pull/222), it's time to refactor the codebase to leverage recently available Rust features for better code ergonomics.

## Solution

This PR provides only cosmetic codebase refactoring. It does no user-facing semantics changes.

Performed refactorings:
- Use [`let`-`else` statement](https://blog.rust-lang.org/2022/11/03/Rust-1.65.0.html#let-else-statements) wherever it fits and simplifies code.
- Use [`matches!()` macro](https://blog.rust-lang.org/2020/03/12/Rust-1.42.html#matches) instead of `match` expression where it improves ergonomics.
- Use [`bool::then()`](https://doc.rust-lang.org/stable/std/primitive.bool.html#method.then)/[`bool::then_some()`](https://doc.rust-lang.org/stable/std/primitive.bool.html#method.then_some) wherever it simplifies code by eliminating `else { None }` branch.
- Use [short formatting syntax](https://blog.rust-lang.org/2022/01/13/Rust-1.58.0.html#captured-identifiers-in-format-strings) where possible (but without changing docs and tests for our formatting macros).
- Leverage full ergonomics of [`format_ident!()` macro](https://docs.rs/quote/1.0/quote/macro.format_ident.html) and remove raw `syn::Ident::new()` usages.
- Prefer `.into()` conversion over `.to_string()`/`.to_owned()` for `&str` where destination type is known to be `String` exactly (it's more future-proof this way, since we don't need to change these places if we would switch destination type to `Cow<'_, str>` or `SmartString`, etc).
- Use [`#![warn(clippy::nonstandard_macro_braces)]`](https://rust-lang.github.io/rust-clippy/rust-1.64.0#nonstandard_macro_braces) to require using the same syntax for macro calls anywhere (we were using `quote!()` in some places and `quote! {}` in others).

## Checklist

- [x] Tests are ~~added/~~ updated

[l:1]: /CHANGELOG.md

### DIFF
--- a/clippy.toml
+++ b/clippy.toml
@@ -1,1 +1,21 @@
+# Project configuration for Clippy Rust code linter.
+# See full lints list at:
+# https://rust-lang.github.io/rust-clippy/master/index.html
+
 msrv = "1.65.0"
+standard-macro-braces = [
+    { name = "assert", brace = "(" },
+    { name = "assert_eq", brace = "(" },
+    { name = "assert_ne", brace = "(" },
+    { name = "format", brace = "(" },
+    { name = "format_ident", brace = "(" },
+    { name = "matches", brace = "(" },
+    { name = "panic", brace = "(" },
+    { name = "parse_quote", brace = "{" },
+    { name = "quote", brace = "{" },
+    { name = "quote_spanned", brace = "{" },
+    { name = "todo", brace = "(" },
+    { name = "unimplemented", brace = "(" },
+    { name = "unreachable", brace = "(" },
+    { name = "vec", brace = "[" },
+]
--- a/clippy.toml
+++ b/clippy.toml
@@ -3,13 +3,15 @@
 # https://rust-lang.github.io/rust-clippy/master/index.html
 
 msrv = "1.65.0"
+
+# Ensures consistent bracing for macro calls in the codebase.
+# Extends default settings:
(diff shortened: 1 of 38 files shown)


clippy.toml
```diff
@@ -1 +1,21 @@
+# Project configuration for Clippy Rust code linter.
+# See full lints list at:
+# https://rust-lang.github.io/rust-clippy/master/index.html
+
 msrv = "1.65.0"
+standard-macro-braces = [
```
JelteF, line 6: Could use a comment on what this is achieving as opposed to the default settings.

JelteF: After checking the PR more I realize it's to add rules for `quote` and `parse_quote` mostly. But afaict there's no need to redefine the default ones already provided by clippy, because it will add any user configured ones to the default list: https://github.com/rust-lang/rust-clippy/blob/master/clippy_lints/src/nonstandard_macro_braces.rs#L187-L190

JelteF (approved):
Changes all look good, only thing I would change is the clippy config

clippy.toml
tyranron: @JelteF I've removed default ones from the list, and added a comment with the link to the default list: https://github.com/JelteF/derive_more/pull/226/commits/e4fe06551b366ceac3e1de8dbd94a2ef02c895f8

JelteF (approved):

PR #231, opened 2023-01-23
Fix installing `tomljson` on CI

tyranron:
## Synopsis

`tomljson` fails to install on CI now due to newer Go versions: https://github.com/JelteF/derive_more/actions/runs/3987055224/jobs/6836368442#step:5:11

More info:  https://golang.org/doc/go-get-install-deprecation

## Solution

Try to use `go install` instead, as the error message suggests.

## Checklist

- [x] ~~Documentation is updated~~ (if required)
- [x] ~~Tests are added/updated~~ (if required)
- [x] ~~[CHANGELOG entry][l:1] is added~~ (if required)

[l:1]: /CHANGELOG.md

### DIFF
--- a/ci/test_all_features.sh
+++ b/ci/test_all_features.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-set -euxo pipefail
+set -eux
 
 for feature in $(tomljson Cargo.toml | jq --raw-output '.features | keys[]' | grep -v 'default\|testing-helpers'); do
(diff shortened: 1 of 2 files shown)


ci/test_all_features.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
```
JelteF, line 1: I'm surprised this is needed but it's fine

JelteF (approved):

Issue #223, opened 2022-11-15
Rename CustomDebug


JelteF:
Copied from #202 for a more structured discussion.

Written by @tyranron:

> 
> I would like in future to add more derives which overlap with the ones from `std`. For example, `#[derive(Clone)]` from `std` is very stupid when it comes to generics, because it's blindly pastes `: Clone` trait bound on types parameters rather than field types, thus leads to [tones of squatting](https://github.com/cucumber-rs/cucumber/blob/da1adca9724fa504f2f56d98ad63561bb779f940/src/writer/normalize.rs#L53-L62). We already have `CustomDebug` which overlaps with `#[derive(Debug)]` and cannot be named `Debug` because of implicit `prelude` import. 
> 
> Usually, we have a code style rule to sort any enumerations in alphabetical order, so the proposed naming change is to omit any possible moves in a list of derived traits when we go from `Debug` to `CustomDebug`, but rather to add a suffix in-place to the standard derive:
> - Instead of this:   
>    `#[derive(AsRef, Clone, Copy, Debug, Default, Display, Eq, Error, From, PartialEq)]` -> `#[derive(AsRef, CustomClone, CustomCopy, CustomDebug, CustomDefault, CustomEq, CustomPartialEq, Display, Error, From)]`
> - Have this:
>   `#[derive(AsRef, Clone, Copy, Debug, Display, Eq, Error, From, PartialEq)]` -> `#[derive(AsRef, CloneCustom, CopyCustom, DebugCustom, DefaultCustom, Display, EqCustom, Error, From, PartialEqCustom)]`
> 
> I would also consider some simpler/shorter suffix rather than `Custom`, but cannot invent any good candidate at the moment rather than `Smart` (which is not ideal, btw).
> 
> `#[derive(AsRef, CloneSmart, CopySmart, DebugSmart, DefaultSmart, Display, EqSmart, Error, From, PartialEqSmart)]`

Written by @JelteF 

> 
> Both proposed breaking changes sound good to me. Can you make separate issues for them so that they can be tracked in the 1.0.0 milestone? I see that 2 is already tracked in #217
> 
> Regarding suffix, one other idea that comes to mind is More (as in coming from derive_more). I don't have a strong preference to any of Custom, Smart or More (tbh all of them sound suboptimal, but I don't know a better one)
> 

Written by @tyranron 

[...]

PR #232, opened 2023-01-24
Showcase some possible bugs in Display and DebugCustom

JelteF:
I played around a bit with DebugCustom and Display after and found some
edge cases that are handled strangely. Are these bugs?

This is not meant to be merged as is. But more as an easy way to share some test
cases with (to me) unexpected results.

Related to #223

### DIFF
--- a/tests/display.rs
+++ b/tests/display.rs
@@ -85,7 +85,7 @@ mod structs {
         #[derive(Binary)]
         struct Binary(i32);
 
-        #[derive(Display)]
+        #[derive(Display, DebugCustom)]
         struct Struct {
             field: i32,
         }
@@ -100,6 +100,7 @@ mod structs {
             assert_eq!(Tuple(0).to_string(), "0");
             assert_eq!(format!("{:b}", Binary(10)), "1010");
             assert_eq!(Struct { field: 0 }.to_string(), "0");
+            assert_eq!(format!("{:?}", Struct { field: 0 }), "0");
             assert_eq!(format!("{:o}", Octal { field: 10 }).to_string(), "12");
         }
 
@@ -154,9 +155,16 @@ mod structs {
             #[display("tuple")]
             struct Tuple(i32, i32);
 
-            #[derive(Display)]
+            #[derive(Display, DebugCustom)]
             #[display("struct")]
+            #[debug("Struct: {field1} {field2}")]
             struct Struct {
+                field1: i32,
+                field2: i32,
+            }
+
+            #[derive(Display, DebugCustom)]
+            struct Struct2 {
                 field1: i32,
                 field2: i32,
             }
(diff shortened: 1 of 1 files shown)


tests/display.rs
```diff
@@ -100,6 +100,7 @@ mod structs {
             assert_eq!(Tuple(0).to_string(), "0");
             assert_eq!(format!("{:b}", Binary(10)), "1010");
             assert_eq!(Struct { field: 0 }.to_string(), "0");
+            assert_eq!(format!("{:?}", Struct { field: 0 }), "0");
```
JelteF, line 103: @tyranron Based on your comment here I would call this unexpected: https://github.com/JelteF/derive_more/issues/223#issuecomment-1316015457

It also makes sense in another way, since this is the newtype derive.

tests/display.rs
```diff
@@ -172,6 +180,33 @@ mod structs {
...
+                assert_eq!(
+                    format!(
+                        "{:?}",
+                        Struct {
+                            field1: 0,
+                            field2: 3
+                        }
+                    ),
+                    "Struct: 0 3"
+                );
```
JelteF, line 192: This one makes sense. This is meant as a showcase of what DebugCustom is currently useful for.

tests/display.rs
```diff
@@ -172,6 +180,33 @@ mod structs {
...
+                assert_eq!(
+                    format!(
+                        "{:?}",
+                        Struct2 {
+                            field1: 0,
+                            field2: 3
+                        }
+                    ),
+                    "Struct2"
+                );
```
JelteF, line 209: Should these cases error?

tests/display.rs
tyranron: @JelteF @ilslv hmm... I'd say that the current `DebugCustom` implementation is not useful at all, in the contrast with `Display` and other `Display`-like derives. The problem is that it follows `Display`-like semantics, but in reality this is not what we expect. I'd propose to remake `DebugCustom` in a separate PR to follow its own `Debug` semantics inherited from `std`'s `#[derive(Debug)]` with steroids in form of:
- structural bounding of type parameters (as we do for `Display`-like macros);
- allowing to change how field is rendered granularly with `#[debug("fmt string", fmt, args)]` attribute;
- fields ignoring with `#[debug(ignore)]`.

JelteF: The behaviour you propose sounds good to me.

tests/display.rs
JelteF:
```rust
                assert_eq!(
                    Struct2 {
                        field1: 0,
                        field2: 3
                    }.to_string(),
                    "Struct2"
                );
```
@ilslv @tyranron do we want to reconsider this behaviour?

tyranron: The [described and implemented design](https://github.com/JelteF/derive_more/pull/225#issuecomment-1319892317) misses default multi-field struct case.

@ilslv should we error in this case?

ilslv: @JelteF @tyranron yes, this definitely should produce compile-time error.

tyranron: @ilslv please, make a fix in a separate PR. Thanks!

ilslv: Fixed in https://github.com/JelteF/derive_more/pull/243

tyranron:
Ooops, GitHub has triggered auto-close on a comment link =/

JelteF:
All issues shown here are now fixed

Issue #221, opened 2022-11-14
Use dedicated error type for errors returned by Add, Not and FromStr


JelteF:
For consistency with #220 we should have custom error types for Add, Not and FromStr too

PR #233, opened 2023-01-25
Use dedicated error types for `Add` and `Not` derives (#221)

tyranron:
Part of #221

## Synopsis

See https://github.com/JelteF/derive_more/issues/221#issue-1448099538

## Solution

Adds `ops::UnitError`, `ops::WrongVariantError` and `ops::BinaryError` to `derive_more` crate and uses them in `Add`-like and `Not`-like derives.

Additionally:
- Renames `derive_more::errors` module as `derive_more::convert` to follow `std`/`core` naming conventions.
- Adds `#[doc(hidden)]` attribute on to `new()` methods of created error types. This way we mimic them as not public, thus won't introduce breaking changes in future once we need to modify their private fields.

## Checklist

- [x] ~~Tests are added/updated~~ (if required)

[l:1]: /CHANGELOG.md


JelteF (approved):
Thanks! looks good.

PR #234, opened 2023-01-25
Disable all features by default

### LINKED ISSUE #210: Remove default features for 1.0
This library is growing more and more, and most people only use a few of the provided derives. So I think a better default might be to  enable none of the features by default. It seems especially useful because some of the less commonly used derives depend on other crates again too, like `convert_case`.

@tyranron any opinion on this?

This would obviously be a breaking change, so I'd want to do this for 1.0 if we indeed want to go this way.

JelteF:
Resolves #210

## Synopsis

As discussed in #210 we want to disable all features by default in 1.0.
With the amount of derives that are part of this library the compile
time of the library itself becomes significant. This way we don't have
to worry about increasing people their compile time when adding more
derives.

On my machine compiling the `derive_more-impl` crate took ~4.5 seconds
when enabling all features. But when only enabling a single derive it
took 1.5 seconds.

Enabling all derives is still easy to do, by using the `full` feature.

## Checklist

[l:1]: /CHANGELOG.md

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,7 +73,7 @@ jobs:
       - name: Install minimal dependencies versions
         run: cargo +nightly update -Z minimal-versions
 
-      - run: cargo test --workspace --features testing-helpers
+      - run: cargo test --workspace --features full,testing-helpers
 
   no_std:
     runs-on: ubuntu-latest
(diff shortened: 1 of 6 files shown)


tyranron (approved):
@JelteF LGTM 👍

PR #238, opened 2023-02-06
Fix generics tokenization in `Error` derive macro

zohnannor:
## Synopsis

`#[derive(Error)]` on `enum`s or `struct`s that have *both* `const` generics and generic types...
```rust
enum ErrConstGeneric<const X: usize, E> {
    Inner(E),
}
```
...panics:
```
error: proc-macro derive panicked
   --> tests/error/derives_for_generic_enums_with_source.rs:270:30
    |
270 |     #[derive(Debug, Display, Error)]
    |                              ^^^^^
    |
    = help: message: expected one of: `for`, parentheses, `fn`, `unsafe`, `extern`, identifier, `::`, `<`, square brackets, `*`, `&`, `!`, `impl`, `_`, lifetime
```

## Solution

For some reason, iterating over [`syn::Generics::params`](https://docs.rs/syn/latest/syn/struct.Generics.html#structfield.params) that consists of only `const` generics or only generic types, tokenizes them as identifiers only, but when we mix them, `syn` decides to output `const` generics like they were defined, with `const` keyword, type annotation, etc. So when we are rendering the trait `impl`, adding bounds for an item in `where` clause, we get this:
```rust
#[automatically_derived]
impl<const X: usize, E> ::std::error::Error for ErrConstGeneric<X, E>
where
    E: ::std::fmt::Debug + ::std::fmt::Display + ::std::error::Error + 'static,
    ErrConstGeneric<const X: usize, E>: ::std::fmt::Debug + ::std::fmt::Display, // obv wrong
{
    fn source(&self) -> Option<&(dyn ::std::error::Error + 'static)> {
        match self {
            ErrConstGeneric::Inner(source) => {
                Some(source as &(dyn ::std::error::Error + 'static))
            }
        }
    }
}
```

Use `syn`'s built-in [`Generics::split_for_impl`](https://docs.rs/syn/latest/syn/struct.Generics.html#method.split_for_impl) to obtain a [type](https://docs.rs/syn/latest/syn/struct.TypeGenerics.html) that properly tokenizes all generic parameters.

As for the tests, I've added a "compile fail" tests for `Error` derive macro on generic `enum`s and `struct`s with combinations of different generics. AFAIU, this happens only in `Error` macro, tell me if I should add tests for some others.

This should be listed in the changelog under `Fixes` section, am I right?

## Checklist

[l:1]: /CHANGELOG.md
[`syn`]: https://docs.rs/syn


impl/src/error.rs
```diff
@@ -48,20 +48,21 @@ pub fn expand(
...
+            }
         }
     });
 
     let mut generics = generics.clone();
 
     if !type_params.is_empty() {
-        let generic_parameters = generics.params.iter();
+        // delegate generic parameters tokenization to `syn`
+        let (_, generic_parameters, _) = generics.split_for_impl();
```
JelteF, line 60: nit: let's make the variable naming consistent with the rest of the codebase

```suggestion
        let (_, ty_generics, _) = generics.split_for_impl();
```

JelteF (commented):
Thanks for the contribution and nice catch. We use split_for_impl for this exact reason in most other places, but it seems this one was missed. Changelog should indeed be under the Fixes section.

tests/error/derives_for_generic_enums_with_source.rs
```diff
@@ -245,5 +245,53 @@ fn named_variant_ignore_redundant() {
...
+    enum ErrGenericConst<E, const X: usize> {
+        Inner(E),
+    }
+    #[derive(Debug, Display, Error)]
+    enum ErrLifetimeGeneric<'lt: 'static, E> {
+        Inner(&'lt E),
+    }
+
+    #[derive(Debug, Display, Error)]
+    enum ErrLifetimeConst<'lt: 'static, const X: usize> {
```
JelteF, line 285: I think a better place for these tests might be in `tests/generics.rs`. That's where we have tests similar to this to test generics for all derives (that support them). They are currently missing const and lifetime ones though, which I guess is why this error was not caught by them.

zohnannor:
Thank you! I'll make corrections and re-request a review when I can.

tyranron (approved):
@zohnannor thanks!

Issue #244, opened 2023-03-07
`Error` doesn't work with generic


KSXGitHub:
Bug description and steps to reproduce is in this repo: https://github.com/KSXGitHub/derive-more-error-bug-generic

JelteF:
Could you share the actual error? That you get?

KSXGitHub:
```
error: proc-macro derive panicked
 --> src/lib.rs:4:26
  |
4 | #[derive(Debug, Display, Error)]
  |                          ^^^^^
  |
  = help: message: expected `,`

error: could not compile `derive-more-error-bug-generic` due to previous error
```

KSXGitHub:
I also deleted `~/.cargo` and reran `cargo clippy`, same result.

ilslv:
@KSXGitHub try adding `features = ["display", "error"]`

KSXGitHub:
@ilslv Your advice works (with minimal edit to existing code).

Regarding the error message for the old syntax. It is too ambiguous in my opinion: It saids <code>use &#96;"{name}: {source}"&#96; instead</code> but the old syntax already has `"{name}: {source}"`! It took me a while to realize that I need to remove `fmt =`. The error would be better if it tells user what to remove instead of what to include.

ilslv:
@KSXGitHub thanks for your perspective. I've addressed error message in https://github.com/JelteF/derive_more/pull/247

tyranron:
Closing this, as this is already resolved on latest `master` and will be released in 1.0.0.

PR #247, opened 2023-03-08
Clarify legacy attribute syntax errors for `Dispaly`-like and `Debug` macros

ilslv:
Resolves https://github.com/JelteF/derive_more/issues/244#issuecomment-1458625976

## Synopsis

> Regarding the error message for the old syntax. It is too ambiguous in my opinion: It saids use `"{name}: {source}"` instead but the old syntax already has `"{name}: {source}"`! It took me a while to realize that I need to remove `fmt =`. The error would be better if it tells user what to remove instead of what to include.

## Solution

Change error from 

```
error: legacy syntax, use `"Stuff({}): {}", bar` instead
 --> tests/compile_fail/display/legacy_fmt_syntax.rs:2:10
  |
2 | #[display(fmt = "Stuff({}): {}", "bar")]
  |          ^
```

to

```
error: legacy syntax, remove `fmt =` and use `"Stuff({}): {}", bar` instead
 --> tests/compile_fail/display/legacy_fmt_syntax.rs:2:10
  |
2 | #[display(fmt = "Stuff({}): {}", "bar")]
  |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

## Checklist

[l:1]: /CHANGELOG.md

### DIFF
--- a/impl/src/fmt/debug.rs
+++ b/impl/src/fmt/debug.rs
@@ -165,10 +165,12 @@ impl ContainerAttributes {
 
 impl Parse for ContainerAttributes {
     fn parse(input: ParseStream) -> Result<Self> {
-        let error_span = input.span();
+        use proc_macro2::Delimiter::Parenthesis;
 
+        let error_span = input.cursor().group(Parenthesis).map(|(_, span, _)| span);
         let content;
         syn::parenthesized!(content in input);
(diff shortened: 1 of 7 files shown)


tyranron (approved):
@ilslv thanks for the quick reaction!

PR #248, opened 2023-03-16
Redesign `Into` derive macro

ilslv:
## Synopsis

This PR is a part of replacing all attributes having `syn::Meta` syntax to custom parsing.

## Solution

Replace `#[into(types(i32, "&str"))]` with `#[from(i32, &str)]` and add support for deriving multi-field structs and enum variants with `#[from((<tuple>), (<tuple>), ...)]`.

## Checklist

[l:1]: /CHANGELOG.md


JelteF:
It's a bit hard to review this as it also contains all the changes from #241. Could you create a PR on your own where the base branch is the from-attribute branch? That way the diff doesn't contain so much unrelated stuff and it's easier to comment on the actual changes that are made for this one.

impl/doc/into.md
```diff
@@ -1,166 +1,119 @@
...
-One thing to note is that this derive doesn't actually generate an
-implementation for the `Into` trait.
-Instead it derives `From` for the values contained in the struct and thus has an
-indirect implementation of `Into` as recommended by the
-[docs](https://doc.rust-lang.org/core/convert/trait.Into.html).
+it should contain, it allows you to extract the values from the struct. One
+thing to note is that this derive doesn't actually generate an implementation
+for the `Into` trait. Instead, it derives `From` for the values contained in
+the struct and thus has an indirect implementation of `Into` as
+[recommended by the docs][1].
```
JelteF, line 9:
```suggestion
[recommended by the Rust docs for `Into`][1].
```

impl/doc/into.md
```diff
@@ -1,166 +1,119 @@
...
+#[derive(Debug, Into, PartialEq)]
+struct Int(i32);
 
-## Tuple structs
+assert_eq!(2, Int(2).into());
+```
 
-When deriving `Into` for a tuple struct with a single field (i.e. a newtype) like this:
+For structs that have multiple fields `.into()` extract tuple containing the
+desired content for each field.
```
JelteF, line 28:
```suggestion
For structs that have multiple fields `.into()` extracts a tuple containing the
desired content for each field.
```

impl/doc/into.md
```diff
@@ -1,166 +1,119 @@
...
-        (original.0)
-    }
-}
+assert_eq!((1, 2), Point(1, 2).into());
 ```
 
-The behaviour is a bit different when deriving for a struct with multiple
-fields, since it returns a tuple. For instance when deriving for a tuple struct
-with two fields like this:
+To specify concrete types to derive convert into use `#[into(<types>)]`.
```
JelteF, line 39:
```suggestion
Use `#[into(<types>)]` to specify the concrete types that you want to convert the type into.
```

impl/doc/into.md
```diff
@@ -1,166 +1,119 @@
...
 ```rust
 # use derive_more::Into;
 #
-#[derive(Into)]
-struct Point1D {
-    x: i32,
-}
-```
+#[derive(Debug, Into, PartialEq)]
+#[into(owned, ref(i32), ref_mut)]
```
JelteF, line 71: So `into(owned)` is equivalent to `into(types({current_type}))`?

JelteF: I think a better example would be to use `ref(i64)`

JelteF: hmm, actually what's the difference between `ref(i64)` and simply using `&i64`?

JelteF:
@ilslv @tyranron now that #241 is merged, let's rebase this one. That should make it much easier to review.

ilslv:
@JelteF I was planning to do it next week 👍

impl/doc/into.md
tyranron: @JelteF here `ref`/`ref_mut` control `Self` type representation, while `&i64` controls target type.

`ref(i64)` means `impl Into<&i64> for &MyType`, while `&i64` means only `impl Into<&i64> for MyType`.

tyranron (approved):
@ilslv I've unified some repeated code (in `utils` module you've referred to `from` via `cfg` clause, but no actual use has been introduced), and corrected docs a llitle bit. Everything else seem nice enough.

@JelteF would you like to take a look, or we can merge it?

JelteF (approved):

PR #254, opened 2023-03-29
Make CHANGELOG.md adhere much more closely to the Keep A Changelog format

### LINKED ISSUE #253: `CHANGELOG.md` does not adhere completely to the Keep A Changelog format
Hi! I was looking through some crates on crates.io, and I happened to chance upon the `CHANGELOG.md` file within this repository.

I noticed that it didn't quite adhere properly to the Keep A Changelog format - despite claiming otherwise in its opening lines. For example:

* The header is `Change Log` instead of `Changelog` (pedantic, I know, but it's good to be consistent c: )
* The `1.0.0 - 2023-??-??` is clearly representative of an `Unreleased` section - so it should be labeled as such.
* The various sections under each version ("Breaking Changes", "New Features", "Improvements", "Fixes") do not match the "Types of changes" that Keep A Changelog proposes to use.

To be clear, I'm being pedantic with all this. The changelog is actually fine as-is! Importantly, it's kept consistent, so it's not as if it's hard to read.

But if you'd allow me, I'd gladly go through and rewrite (or rather, rearrange) `CHANGELOG.md` to fix the things listed above. Would that be okay?

LikeLakers2:
Resolves #253.

Some specific, noteworthy changes:
* The "Internal changes" sections under 0.7.1 and 0.9.0 have been removed - users of your library don't need (nor do they care) about internal details, so long as it works.
* The section for 0.99.4 was retitled to include `[YANKED]`.
* Certain changes, such as the MSRV changes, or the change about all features except `std` being disabled by default, were reworded. This is either for clarity, or to make them fit better as changelog entries.

And some notes:
* Since all breaking changes (that I know of) were in 0.x.0 sections (and the unreleased/1.0.0 section), I did not edit them to specify "BREAKING CHANGE." Users that know about SemVer (which we can safely assume to be 99% of Rust users, due to its prevalence) will know that these versions could contain breaking changes.
* Keep A Changelog prefers that version headers link to a diff of the changes made (you can see this on Keep A Changelog's own changelog). As you haven't been using the tags functionality, I didn't bother to link those diffs.

I may perform a second pass sometime after this one is merged. I've gotten most of what I've noticed out of the way, but I still feel I could make it a little better (for example, the diff links).

## Checklist

- ~~[ ] Documentation is updated (if required)~~
- ~~[ ] Tests are added/updated (if required)~~
- ~~[ ] [CHANGELOG entry][l:1] is added (if required)~~

(I have not included a CHANGELOG entry as this does not apply to the library itself.)

[l:1]: /CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-# Change Log
+# Changelog
 
 All notable changes to this project will be documented in this file.
 
@@ -6,14 +6,24 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 
-## 1.0.0 - 2023-??-??
+## Unreleased
 
-### Breaking changes
+### Added
 
-- The minimum supported rust version is updated to Rust 1.65.
-- The `From` derive doesn't derive `From<()>` for enum variants without any
-  fields anymore. This feature was removed because it was considered useless in
-  practice.
+- Add the `std` feature which should be disabled in `no_std` environments.
+- Add support captured identifiers in `Display` derives. So now you can use:
+  `#[display(fmt = "Prefix: {field}")]` instead of needing to use
+  `#[display(fmt = "Prefix: {}", field)]`
+- Add `FromStr` derive support for enums that contain variants without fields.
+  If you pass the name of the variant to `from_str` it will create the matching
+  variant.
+
+### Changed
+
+- The MSRV is now Rust 1.65.
+- All Cargo features, except `std`, are now disabled by default. The `full`
+  feature can be used to get the old behavior of supporting all possible
+  derives.
 - The `TryFrom`, `Add`, `Sub`, `BitAnd`, `BitOr`, `BitXor`, `Not` and `Neg`
   derives now return a dedicated error type instead of a `&'static str` on
   error.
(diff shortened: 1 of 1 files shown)


JelteF:
I took a quick look and most of the changes look reasonable. The main change I disagree with is the removal of the "breaking changes" section. This is the main thing people should takeaway when reading the changelog. So let's keep that section and let's keep it at the top of each version that has them.

> Users that know about SemVer (which we can safely assume to be 99% of Rust users, due to its prevalence) will know that these versions could contain breaking changes.

While people indeed will know that there are breaking changes, the changelog is there to show what changes are actually breaking. It's important not to miss any, by having them "hidden" between many other less interesting changes.

There's also a pretty long ongoing discussion on the "keep a changelog" repo regarding this: https://github.com/olivierlacan/keep-a-changelog/discussions/480

LikeLakers2:
I would argue against keeping it - Labeling a change specifically as "Breaking change" provides no benefit over labeling it as "Added", "Removed", or the like. Additionally, all versions of `derive_more` with breaking changes are either 0.x -> 0.y versions, or the 1.0.0 version, so anyone who understands SemVer (again, 99% of Rust users) will automatically understand that those versions could have breaking changes (documented or not).

I can still re-add it if you still want it, but I *really* don't see a purpose to keeping it.

JelteF:
> so there is no point to calling them "Breaking" when people will assume all changes within those sections will be breaking in some manner.

But that's not the case. Not all of the changes in a backwards incompatible release are breaking. Many of the changes are not. So calling out the specific changes that might break something by updating is very useful to users. Which is also why this section is at the top of the changelog entry, because it's likely the most important to users. Instead of as suggested having a "Removed" section at the bottom. 

So yes, please re-add. People in the discussion I linked in my previous comment use the same approach.

LikeLakers2:
> But that's not the case. Not all of the changes in a backwards incompatible release are breaking.

Man, that's not what I was saying at all. I was saying that many people assume as such, since breaking changes are only ever allowed on 0.x -> 0.y or x.0 -> y.0 releases. Whether or not it's the case is wholly irrelevant here.

Look, just... just take my commit and edit it to your liking. I don't need credit or the like, and I don't particularly care for the copyright of a commit like this anyways, so you can do with it as you please - including re-adding the "Breaking changes" section.

JelteF:
I merged #256 just now, which is heavily based on this PR. Thanks for your help.

PR #265, opened 2023-06-21
Support multiple attribute occurrences

zohnannor:
## Synopsis

It is impossible to feature-gate additional type conversions in `From`, `Into` derive macros. Code such as:

```rust
#[derive(From)]
#[from(types(u8, u16))]
#[cfg_attr(feature = "myfeature", from(types(u32)))]
struct MyIntExplicit(u64);
```

will fail to compile with `Only a single attribute is allowed` when `myfeature` is enabled.

## Solution

Instead of err-ing on second occurrence of the attribute when building `MetaInfo`, just proceed and collect all additional attributes into it. Anyway user will see the error message if they'll try to provide types to attributes, expansions of which will result in conflicting implementations.

As I understand the inner working of the macro, that's all there is to it to support multiple attributes?

## Checklist

- [ ] [CHANGELOG entry][l:1] is added (if required)

[l:1]: /CHANGELOG.md

### DIFF
--- a/impl/src/into.rs
+++ b/impl/src/into.rs
@@ -49,22 +49,22 @@ pub fn expand(input: &DeriveInput, trait_name: &'static str) -> Result<TokenStre
             let into_types: Vec<_> = field_types
                 .iter()
                 .map(|field_type| {
-                    // No, `.unwrap_or()` won't work here, because we use different types.
-                    if let Some(type_) = explicit_type {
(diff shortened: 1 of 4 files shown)


zohnannor:
Seems like this error will be fixed in #241, #248 PRs.

tyranron:
@zohnannor seems to be resolved in the latest beta releases. Feel free to reopen if something more is required.

PR #273, opened 2023-07-15
Re-export traits

### LINKED ISSUE #271: re-export traits together with the derive
This seems quite nice for traits that are not in the `prelude`. Often if you're deriving things you also want to use them.

JelteF:
Re-export traits from `std` for all our derives. This way people don't need to import traits manually when they want to reference them somewhere, such as bounds.

Fixes #271


JelteF:
I created this issue on the internals forum: https://internals.rust-lang.org/t/use-ing-only-a-trait-not-a-derive-with-the-same-name/19153

tyranron:
@JelteF could you specify an example where it causes trouble? I don't understand where the conflict appears exactly.

tyranron:
@JelteF ok, investigated this a liitle bit more and understood the problem. Seems like [this answer does the trick](https://internals.rust-lang.org/t/use-ing-only-a-trait-not-a-derive-with-the-same-name/19153/2) even if `std` will ad new derives in the future.

Will commit fix shortly.

tyranron:
It also makes traits visible in docs, but I do prefer this. It's much more confusing to not see traits in docs while they're imported, actually.

src/lib.rs
```diff
@@ -77,18 +77,106 @@ mod vendor;
...
+
+#[cfg(feature = "as_mut")]
+#[doc(hidden)]
+pub use core::convert::AsMut;
+
+// XXX: Uncommenting this causes our own derive to not be visible anymore, because the derive
+// from std takes precedence somehow.
+// #[cfg(feature = "debug")]
+// #[doc(hidden)]
+// pub use core::fmt::Debug;
```
JelteF, line 109: To be clear, this was the example that caused issues.

JelteF:
> Seems like [this answer does the trick](https://internals.rust-lang.org/t/use-ing-only-a-trait-not-a-derive-with-the-same-name/19153/2) even if `std` will ad new derives in the future.

Yeah, that's a clever trick. Using that seems fine to me.

> It also makes traits visible in docs, but I do prefer this. It's much more confusing to not see traits in docs while they're imported, actually.

On one hand it's nice, on the other hand it clutters the derive_more docs quite a bit. Especially since the traits show up before the derives I think.

tyranron:
@JelteF such refactoring also discovered an `"iterator"` Cargo feature, which simply did nothing (no actual derive existed for it).

tyranron (dismissed):
@JelteF please, take a look, and if everything is OK, let's merge it!

src/lib.rs
```diff
@@ -47,52 +47,171 @@
...
+    pub use crate::fmt::{debug_tuple, DebugTuple};
+
+    #[cfg(feature = "error")]
+    pub use crate::vendor::thiserror::aserror::AsDynError;
+}
+
+/// Module containing macro definitions only, without corresponding traits.
+///
+/// Use it in your import paths, if you don't want to import traits, but only macros.
+pub mod macros {
```
JelteF, line 63: I cannot really think of a reasonable case when users would want this. I think it only clutters the docs. If people really want to do this, they could import a derive directly from `derive_more_impl`. So I feel we should remove this module. What do you think?

src/lib.rs
```diff
@@ -47,52 +47,171 @@
...
+        mod $new_module_name {
+            pub use $module::{$($traits),*};
+        }
 
-#[cfg(feature = "from_str")]
-mod r#str;
-#[cfg(feature = "from_str")]
-pub use self::r#str::FromStrError;
+        #[cfg(feature = $feature)]
+        #[doc(hidden)]
```
JelteF, line 112: I really disliked how the standard library trait listing was taking up a third of the front page of the documentation. Especially beacuse they came before the documentation of the actual derives. So I hid them again and I added a small snippet to the README to say that we re-export all traits.

src/lib.rs
```diff
@@ -47,52 +47,171 @@
...
-#[cfg(any(feature = "add", feature = "not"))]
-pub mod ops;
+// When re-exporting traits from std we need to do a pretty crazy trick, because we ONLY want
+// to re-export the traits and not derives that are called the same in the std module,
+// because those would conflict with our own. The way we do this is by first importing both
+// the trait and possible derive into a separate module and re-export them. Then we wildcard import
+// all the things from that module into the main module, but we also import our own derive by its
+// exact name. Due to the way wildcard imports work in rust, that results in our own derive taking
+// precedence over any derive from std.
+macro_rules! re_export_traits((
```
JelteF, line 104: I created a simple macro to generate the boilerplate code.

impl/src/lib.rs
```diff
@@ -108,27 +108,12 @@ macro_rules! create_derive(
     }
 );
 
-create_derive!("from", from, From, from_derive, from);
```
JelteF, line 111: I reorderd these too, to match the ordering in the other files. I think this orderering was simply the order in which the derives were added.

JelteF (commented):
@tyranron I made some more changes. I think it's almost ready now, but I would like to remove the `macros` module.

PR #275, opened 2023-07-24
Fix docs on docs.rs

tyranron:
## Synopsis

We have a [broken build on docs.rs](https://docs.rs/crate/derive_more/1.0.0-beta.1/builds/868214) for the [latest release](https://github.com/JelteF/derive_more/releases/tag/v1.0.0-beta.1).

## Solution

- Use `full` feature on docs.rs to build crate docs.
- Add the `full` feature to `derive_more-impl` crate too, just for the same purpose.
- Correct CI to check building docs exactly as they're built on docs.rs.

## Additionally

Switch to `rustdoc::all` lints.

## Checklist

- [x] Tests are added/updated
- [x] ~~[CHANGELOG entry][l:1] is added (if required)~~

[l:1]: /CHANGELOG.md

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -43,7 +43,7 @@
 #![cfg_attr(not(feature = "std"), no_std)]
 #![cfg_attr(all(not(feature = "std"), feature = "error"), feature(error_in_core))]
 #![cfg_attr(docsrs, feature(doc_auto_cfg))]
-#![deny(rustdoc::broken_intra_doc_links, rustdoc::private_intra_doc_links)]
+#![deny(rustdoc::all)]
 #![forbid(non_ascii_idents, unsafe_code)]
 #![warn(clippy::nonstandard_macro_braces)]
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -43,7 +43,7 @@
 #![cfg_attr(not(feature = "std"), no_std)]
 #![cfg_attr(all(not(feature = "std"), feature = "error"), feature(error_in_core))]
 #![cfg_attr(docsrs, feature(doc_auto_cfg))]
-#![deny(rustdoc::all)]
+#![cfg_attr(any(not(docsrs), ci), deny(rustdoc::all))]
 #![forbid(non_ascii_idents, unsafe_code)]
 #![warn(clippy::nonstandard_macro_braces)]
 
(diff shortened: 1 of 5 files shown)


src/lib.rs
```diff
@@ -43,7 +43,7 @@
 #![cfg_attr(not(feature = "std"), no_std)]
 #![cfg_attr(all(not(feature = "std"), feature = "error"), feature(error_in_core))]
 #![cfg_attr(docsrs, feature(doc_auto_cfg))]
-#![deny(rustdoc::broken_intra_doc_links, rustdoc::private_intra_doc_links)]
+#![deny(rustdoc::all)]
```
JelteF, line 46: I feel like we want to do this for dev and CI, but not when docs.rs is actually building the docs. Otherwise it might fail to do so if some new lint gets added. Let's instead add `-D rustdoc::all` to the `RUSTDOCFLAGS` in CI.

tyranron: @JelteF 

> otherwise it might fail to do so if some new lint gets added

How is it possible? Build on docs.rs is done right after CI passed and crate is released.

> Let's instead add `-D rustdoc::all` to the `RUSTDOCFLAGS` in CI. (we should double check that it works)

Don't really like the idea of additional flags here and there.

Maybe `#![cfg_attr(not(docsrs), deny(rustdoc::all))]` would do the trick?

JelteF: > How is it possible? Build on docs.rs is done right after CI passed and crate is released.

I'm pretty sure docs.rs might rebuild docs later again after a new release of rustdoc (e.g. to get new styling).

> Maybe `#![cfg_attr(not(docsrs), deny(rustdoc::all))]` would do the trick?

I guess that would work, but would want to re-enable the the errors for CI so probably something like:
```
#![cfg_attr(any(not(docsrs), ci), deny(rustdoc::all))]
```

And then adding `--cfg ci` to the RUSTDOCFLAGS in CI.

tyranron: @JelteF 

> I'm pretty sure docs.rs might rebuild docs later again after a new release of rustdoc (e.g. to get new styling).

Never heard of that. In my experience, once docs are built, they remain as they were built. Having all the old style forever. You can see the old style on old crates, like <https://docs.rs/iron/0.4.0>.

> but would want to re-enable the the errors for CI so probably something like

👍

JelteF (approved):
I'll try to publish a new beta release soon after this is merged so we have docs for the beta.

PR #282, opened 2023-07-26
Re-export errors from ops and unexport ops module

JelteF:
Resolves 

Part of 

Required for 

Requires 

Related to 

## Synopsis

Re-export error types from the ops module from the root of the crate. And
unexport the ops module itself.

This also moves errors that are specific to add to a separate file.
Otherwise rustdoc would show them to be included for the features:
"(add or not) and add", which is obviously equivalent to add but reads
super weird.

## Checklist

- [ ] ~Documentation is updated (if required)~
- [ ] ~Tests are added/updated (if required)~
- [ ] ~[CHANGELOG entry][l:1] is added (if required)~

[l:1]: /CHANGELOG.md


tyranron (approved):
@JelteF I've moved `add` module back to `ops` as `ops::add` one. This seems more logical to me since `add` uses types from `ops`, and in `std` they're all united under `ops` module too.

Please, take a look and merge it, if it's OK for you.

tyranron (approved):
@JelteF done.

PR #284, opened 2023-08-08
Fix IntoIterator derived bounds (#208)

### LINKED ISSUE #208: IntoIterator for generic type includes unnecessary trait bounds
I'm trying to derive `IntoIterator` for a custom struct that wraps over [Bevy](http://bevyengine.org)'s `Query` type, among other things:

```rust
#[derive(SystemParam, Deref, DerefMut, IntoIterator)]
pub struct AnimationControlsQuery<'w, 's, K, F = ()>
    where K: Copy + Eq + Hash + Send + Sync + TypeUuid + 'static,
          F: WorldQuery + ReadOnlyWorldQuery + 'static
{
    #[deref]
    #[deref_mut]
    #[into_iterator(ref, ref_mut)]
    query: Query<'w, 's, AnimationControls<K>, F>,
    animation_assets: Res<'w, Assets<Animation>>,
    animation_set_assets: Res<'w, Assets<AnimationSet<K>>>,
}
```

with the intended usage being roughly:

```rust
fn update_enemy_animations(
    mut animation_controls: AnimationsControlsQuery<EnemyAnimation, With<Enemy>>,
) {
    for mut anim_ctls in &mut animation_controls {
        // do something with anim_ctls which is of type &mut AnimationControls<EnemyAnimation>
    }
}
```

However, as per `cargo expand`, this generates the `IntoIterator` impls for `&'a AnimationControlsQuery<'w, 's, K, F>` and `&'a mut AnimationControlsQuery<'w, 's, K, F>` that have the following trait bounds:

```rust
impl<'w, 's, '__deriveMoreLifetime, K, F> ::core::iter::IntoIterator
    for &'__deriveMoreLifetime AnimationControlsQuery<'w, 's, K, F>
    where
        &'__deriveMoreLifetime K: ::core::iter::IntoIterator,
        &'__deriveMoreLifetime F: ::core::iter::IntoIterator,
        K: Copy + Eq + Hash + Send + Sync + TypeUuid + 'static,
        F: WorldQuery + ReadOnlyWorldQuery + 'static,
    {
        // omitted for brevity
    }
// and analogously for the mut version
```

The extra bounds on `K` and `F` are unnecessary and incorrect, rendering the impls unusable. Neither of those generic parameters are intended to be iterators/iterables themselves, and they aren't even used inside the generated `into_iter` for anything other than for specifying the exact `Query<...>` type and its `into_iter` method to delegate to:

```rust
        #[inline]
        fn into_iter(self) -> Self::IntoIter {
            <&'__deriveMoreLifetime Query<
                'w,
                's,
                AnimationControls<K>,
                F,
            > as ::core::iter::IntoIterator>::into_iter(&self.query)
        }
// similarly for the ::Item and ::IntoIter associated types
```

Thus I think those bounds should be removed by default, and perhaps an `#[into_iterator]` attribute param could be introduced for the use cases that do need them (similar to Serde's `#[serde(bound = "T: MyTrait")]`).

MegaBluejay:
Resolves #208 

## Synopsis

For generic structs, some derived IntoIterator bounds are based on the struct's type parameters, rather than the field used in the implementation.

## Solution

Require only the type of the field used to be `IntoIterator`

## Checklist

## Note

I don't think this is a breaking change, since no program where the newly added bounds aren't satisfied would have compiled.

[l:1]: /CHANGELOG.md

### DIFF
--- a/impl/src/into_iterator.rs
+++ b/impl/src/into_iterator.rs
@@ -1,5 +1,5 @@
 use crate::utils::{
-    add_extra_generic_param, add_extra_ty_param_bound_ref, SingleFieldData, State,
+    add_extra_generic_param, add_extra_where_clauses, SingleFieldData, State,
 };
 use proc_macro2::TokenStream;
 use quote::{quote, ToTokens};
@@ -26,16 +26,19 @@ pub fn expand(input: &DeriveInput, trait_name: &'static str) -> Result<TokenStre
         let reference_with_lifetime = ref_type.reference_with_lifetime();
 
         let generics_impl;
-        let generics =
-            add_extra_ty_param_bound_ref(&input.generics, trait_path, ref_type);
-        let (_, ty_generics, where_clause) = generics.split_for_impl();
         let (impl_generics, _, _) = if ref_type.is_ref() {
-            generics_impl = add_extra_generic_param(&generics, lifetime.clone());
+            generics_impl = add_extra_generic_param(&input.generics, lifetime.clone());
             generics_impl.split_for_impl()
         } else {
-            generics.split_for_impl()
+            input.generics.split_for_impl()
         };
-        // let generics = add_extra_ty_param_bound(&input.generics, trait_path);
+
+        let generics = add_extra_where_clauses(
+            &input.generics,
+            quote! { where #reference_with_lifetime #field_type: #trait_path },
+        );
+        let (_, ty_generics, where_clause) = generics.split_for_impl();
+
         let casted_trait = &quote! {
             <#reference_with_lifetime #field_type as #trait_path>
         };
--- a/tests/into_iterator.rs
+++ b/tests/into_iterator.rs
@@ -7,28 +7,13 @@ extern crate alloc;
 
 #[cfg(not(feature = "std"))]
 use alloc::vec::Vec;
-
 use core::fmt::Debug;
 
 use derive_more::IntoIterator;
 
-fn test_into_iter<T: PartialEq + Debug, I: IntoIterator<Item = T>>(
-    iter: I,
-    vals: &[T],
-) {
+#[track_caller]
+fn assert_iter<T: PartialEq + Debug, I: IntoIterator<Item = T>>(iter: I, vals: &[T]) {
     assert_eq!(iter.into_iter().collect::<Vec<_>>(), vals);
-}
-
-fn test_into_iter_all<T, I>(mut iter: I, mut vals: Vec<T>)
-where
-    T: PartialEq + Debug,
-    I: IntoIterator<Item = T>,
-    for<'a> &'a I: IntoIterator<Item = &'a T>,
-    for<'a> &'a mut I: IntoIterator<Item = &'a mut T>,
-{
-    test_into_iter(&mut iter, &vals.iter_mut().collect::<Vec<_>>());
-    test_into_iter(&iter, &vals.iter().collect::<Vec<_>>());
-    test_into_iter(iter, &vals);
 }
 
 #[derive(IntoIterator)]
@@ -37,9 +22,12 @@ struct MyVec(Vec<i32>);
 
 #[test]
 fn tuple_single() {
-    let numbers = vec![1, 2, 3];
+    let mut vals = vec![1, 2, 3];
+    let mut iter = MyVec(vals.clone());
 
-    test_into_iter_all(MyVec(numbers.clone()), numbers);
+    assert_iter(&mut iter, &vals.iter_mut().collect::<Vec<_>>());
+    assert_iter(&iter, &vals.iter().collect::<Vec<_>>());
+    assert_iter(iter, &vals);
 }
 
 #[derive(IntoIterator)]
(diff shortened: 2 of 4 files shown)


tests/into_iterator.rs
```diff
@@ -45,3 +45,25 @@ impl ::core::iter::IntoIterator for Numbers3 {
         <Vec<i32> as ::core::iter::IntoIterator>::into_iter(self.numbers)
     }
 }
+
+#[derive(IntoIterator)]
+struct Generic1<T> {
+    #[into_iterator(owned, ref, ref_mut)]
+    items: Vec<T>,
+}
```
tyranron, line 53: @MegaBluejay it would be nice to add actual tests for values, to ensure that the generated code produces correct `Iterator`s.

For already existing tests too.

tests/into_iterator.rs
```diff
@@ -45,3 +45,25 @@ impl ::core::iter::IntoIterator for Numbers3 {
...
+{
+    #[into_iterator(owned, ref, ref_mut)]
+    items: Vec<T>,
+    useless: &'a U,
+}
+
+#[derive(IntoIterator)]
+struct Generic3<'a, 'b, T> {
+    #[into_iterator(owned, ref, ref_mut)]
+    items: &'a mut Vec<&'b mut T>,
```
tyranron, line 68: @MegaBluejay good 👍

It's also worth to add the tests which check just `#[into_iterator]` attribute (without `owned`/`ref` inside) against generics.

impl/src/into_iterator.rs
```diff
@@ -26,16 +26,19 @@ pub fn expand(input: &DeriveInput, trait_name: &'static str) -> Result<TokenStre
...
             generics_impl.split_for_impl()
         } else {
-            generics.split_for_impl()
+            input.generics.split_for_impl()
         };
-        // let generics = add_extra_ty_param_bound(&input.generics, trait_path);
+
+        let generics = add_extra_where_clauses(
+            &input.generics,
+            quote! { where #reference_with_lifetime #field_type: #trait_path },
```
tyranron, line 38: @MegaBluejay how about reworking the whole `derive(IntoIterator)` macro implementation in the style the `derive(Into)` macro is implemented: defined typed attribute and expansion structure?

It also would be nice to add `tests/compile_fail/into_iter`.

And we may rename the whole PR as "`IntoIterator` macro rework" or something.

tyranron (changes requested):

impl/src/into_iterator.rs
JelteF: Personally, to keep reviewing easy, I prefer to keep refactoring separate from bugfixes/features, if possible. PRs that do refactoring are usually hard to review, because they change lots of code. Reviewing is made a lot easier if you can assume that the changed code should do the same. Once you also start adding features/bugfixes to such a changeset it becomes really hard to understand if a change in behaviour was unintended, or if it was necessary to for the feature/bugfix.

Adding tests that are currently missing always appreciated though. I don't think it's necessary to block the PR on that though. The change seems simple enough, that I wouldn't expect it to introduce breakage.

tests/into_iterator.rs
JelteF: Especially for the new tests I think this would be great. If you don't mind doing it for the old tests too that is very much appreciated.

impl/src/into_iterator.rs
tyranron: @MegaBluejay well then... let's make a rework in a separate PR. For now, only add a rich set of tests.

tests/into_iterator.rs
MegaBluejay: > It's also worth to add the tests which check just `#[into_iterator]` attribute (without `owned`/`ref` inside) against generics.

Done

tyranron: @MegaBluejay yup, this is out-of-scope. So, recompose the test the way it makes sense.

tyranron (changes requested):
Tests a re still failing.

tests/into_iterator.rs
```diff
@@ -2,23 +2,64 @@
...
 
+use core::fmt::Debug;
+
 use derive_more::IntoIterator;
 
+fn test_into_iter<T: PartialEq + Debug, I: IntoIterator<Item = T>>(
+    iter: I,
+    vals: &[T],
+) {
+    assert_eq!(iter.into_iter().collect::<Vec<_>>(), vals);
```
tyranron, line 16: @MegaBluejay such unification is not good for tests, as doesn't allow us to understand well which test case is failing exactly: 
```
---- generic stdout ----
thread 'generic' panicked at 'assertion failed: `(left == right)`
  left: `[1, 2, 3]`,
 right: `[]`', tests/into_iterator.rs:18:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- generic_bounds stdout ----
thread 'generic_bounds' panicked at 'assertion failed: `(left == right)`
  left: `[1, 2, 3]`,
 right: `[]`', tests/into_iterator.rs:18:5

---- generic_refs stdout ----
thread 'generic_refs' panicked at 'assertion failed: `(left == right)`
  left: `[1, 2, 3]`,
 right: `[]`', tests/into_iterator.rs:18:5

---- generic_owned stdout ----
thread 'generic_owned' panicked at 'assertion failed: `(left == right)`
  left: `[1, 2, 3]`,
 right: `[]`', tests/into_iterator.rs:18:5

---- named_single stdout ----
thread 'named_single' panicked at 'assertion failed: `(left == right)`
  left: `[1, 2, 3]`,
 right: `[]`', tests/into_iterator.rs:18:5

---- tuple_single stdout ----
thread 'tuple_single' panicked at 'assertion failed: `(left == right)`
  left: `[1, 2, 3]`,
 right: `[]`', tests/into_iterator.rs:18:5
```

`#[track_caller]` may be helpful here to some degree, but not when multiple assertions are united in the same helper function.

tyranron (approved):

Issue #285, opened 2023-08-08
AsRef restricted impls


MegaBluejay:
Currently there are 2 options for `AsRef` derives:

- the default behaviour, where only implementations for field types are generated
- `[as_ref(forward)]`, which generates a generic impl for all types the field can be converted to

It would be useful to be able to list types for implementation explicitly, like with [From](https://docs.rs/derive_more/1.0.0-beta.3/derive_more/derive.From.html)

For example

```rust
#[derive(AsRef)]
struct Test {
    #[as_ref(str)]
    str_field: String,
    #[as_ref([u8])]
    u8_field: Vec<u8>,
}
```

This wouldn't work with `[as_ref(forward)]`, since we'd have conflicting implementations for `AsRef<[u8]>`.

Issue #123, opened 2020-04-10
`derive(AsRef)` should also derive identity impl.


wdanilo:
Hi! I believe that:
```rust
#[derive(AsRef)]
struct Foo { t:i32 }
```

should generate identity `AsRef` impl in addition to what it generates now:

```rust
impl AsRef<Foo> for Foo {
    fn as_ref(&self) -> &Self {
        self
    }
}
```

Making these instances by hand is cumbersome and they should be generated automatically.

JelteF:
I'm not entirely sure when you would actually want this, but if you can give some useful example I'm definitely open to accepting PRs for this. Should be behind a attribute flag though, e.g. `as_ref(Self)`

Iizuki:
This would be a useful workaround for the current type system limitations: https://doc.rust-lang.org/std/convert/trait.AsRef.html#reflexivity

The docs state that such an identity implementation would ideally be automatically generated, but that's just not possible for technical reasons currently.

PR #286, opened 2023-08-09
Refactor `AsRef` and `AsMut` derives

MegaBluejay:
## Synopsis

This PR is a part of replacing all attributes having `syn::Meta` syntax to custom parsing, similarly to #241, #248 

Paves the way for #285, and possibly other enhancements such as an opt-in #123

## Solution

Implement custom attribute parsing without `State`

## Checklist

[l:1]: /CHANGELOG.md


impl/src/as_ref.rs
```diff
@@ -79,3 +76,193 @@ pub fn expand(input: &DeriveInput, trait_name: &'static str) -> Result<TokenStre
...
+        attrs
+            .as_ref()
+            .iter()
+            .filter(|attr| attr.path().is_ident("as_ref"))
+            .try_fold(None, |mut attrs, attr| {
+                let field_attr = attr.parse_args()?;
+                if attrs.replace(field_attr).is_some() {
+                    Err(syn::Error::new(
+                        attr.path().span(),
+                        "only single `#[as_ref(...)]` attribute is allowed here",
```
JelteF, line 95: can you add some basic `compile_fail` tests for the errors you're introducing here?

impl/src/as_ref.rs
```diff
@@ -79,3 +76,193 @@ pub fn expand(input: &DeriveInput, trait_name: &'static str) -> Result<TokenStre
...
+    }
+}
+
+impl Parse for StructAttribute {
+    fn parse(input: ParseStream) -> Result<Self> {
+        input.parse::<syn::Path>().and_then(|path| {
+            if path.is_ident("forward") {
+                Ok(Self::Forward)
+            } else {
+                Err(syn::Error::new(path.span(), "unknown"))
```
JelteF, line 110: this error isn't very descriptive
```suggestion
                Err(syn::Error::new(path.span(), "unknown argument, only `forward` is allowed"))
```

impl/src/as_ref.rs
```diff
@@ -79,3 +76,193 @@ pub fn expand(input: &DeriveInput, trait_name: &'static str) -> Result<TokenStre
...
+            return Ok(Self::AsRef);
+        }
+        attr.parse_args::<syn::Path>().and_then(|p| {
+            if p.is_ident("forward") {
+                return Ok(Self::Forward);
+            }
+            if p.is_ident("ignore") {
+                return Ok(Self::Ignore);
+            }
+            Err(syn::Error::new(p.span(), "unknown argument"))
```
JelteF, line 152:
```suggestion
            Err(syn::Error::new(p.span(), "unknown argument, only `forward` and `ignore` are allowed"))
```

JelteF (commented):
Thanks for the contribution. Just to be clear, this isn't adding any new features yet right? Only doing refactoring to allow future features. (I like that approach, just making sure I understand correctly)

MegaBluejay:
>  Only doing refactoring to allow future features.

Yes, that's correct, this PR is refactoring only

MegaBluejay:
Since they're essentially identical, I think it makes sense to use the same deriving function for `AsRef` and `AsMut`.

I can't think of any hypothetical features that should apply to only one of them, and this would facilitate keeping them in sync.

impl/src/ref_conv.rs
```diff
@@ -0,0 +1,312 @@
...
+            fn #method_ident(& #mutability self) -> #return_types {
+                #bodies
+            }
+        }
+    )*})
+}
+
+struct StructAttr<'a> {
+    args: StructAttrArgs,
+    attr: &'a syn::Attribute,
```
tyranron, line 86: @MegaBluejay the code in this file lacks documentation very much, which makes it much difficult to understand the code. See `fmt` modules as a reference for good documentation.

impl/src/ref_conv.rs
```diff
@@ -0,0 +1,312 @@
...
+        .collect();
+    let bodies = sub_items.iter().map(|i| &i.0);
+    let impl_generics = sub_items.iter().map(|i| &i.1);
+    let where_clauses = sub_items.iter().map(|i| &i.2);
+    let trait_paths = sub_items.iter().map(|i| &i.3);
+    let return_types = sub_items.iter().map(|i| &i.4);
+
+    Ok(quote! {#(
+        #[automatically_derived]
+        impl #impl_generics #trait_paths for #input_type #ty_generics #where_clauses {
```
tyranron, line 75: @MegaBluejay I think, here it would be much better if instead of making multiple iterators for each part to use in final `quote! {}`, we just make an iterator of final `quote! {}`s, so we don't squat with those iterators redudantly.

```rust
field_args.into_iter().map(|....| {
    quote! {
        #[automatically_derived]
        impl #impl_generics #trait_paths for #input_type #ty_generics #where_clauses {
            #[inline]
            fn #method_ident(& #mutability self) -> #return_types {
                #bodies
            }
        }
    }
}).collect()
```

impl/src/ref_conv.rs
```diff
@@ -0,0 +1,312 @@
...
+                 ident,
+             }| {
+                let member = quote! { self.#ident };
+                let field_type = &field.ty;
+                if forward {
+                    let trait_path = quote! { #trait_path<#conv_type> };
+                    let type_where_clauses = quote! {
+                        where #field_type: #trait_path
+                    };
+                    let new_generics = add_where_clauses_for_new_ident(
```
tyranron, line 38: @MegaBluejay I would like to get rid of using this old function from `utils`, so we can rip old stuff totally, in the future. It also contains some logic (like adding `Copy` markers) which is totally irrelevant for these tratis.

tyranron (changes requested):
@MegaBluejay not bad, but needs polishing.

I've restructured modules in the way we have for `fmt` macros.

Also, please look at `tests/as_ref.rs` and `tests/as_mut.rs`, whether it makes sense to extend their test cases to cover more edge cases.

tyranron:
@JelteF the interesting question:

Since `AsRef` and `AsMut`  are effectively the same regarding implementation, should we unite them under a single `as_ref` Cargo feature as we do for `Display`-like derives?

I think it's quite reasonable. May be done as a separate PR, though, after merging this one.

impl/src/ref_conv.rs
MegaBluejay: I kept the more basic functions (`add_extra_where_clauses`, `add_extra_generic_type_param`)

They seem general enough to be keep using everywhere

impl/src/as/mod.rs
```diff
@@ -0,0 +1,371 @@
...
+                    let member = quote! { self.#ident };
+                    let field_type = &field.ty;
+                    let (
+                        body,
+                        impl_generics,
+                        where_clause,
+                        trait_path,
+                        return_type
+                    ) = if forward {
+                        let trait_path = quote! { #trait_path<#conv_type> };
```
JelteF, line 55: In general I prefer to have a bit less nested code than this. If it's reasonably possible to un-indent this somehow I would like that. Feel free to ignore this comment though if it requires major changes.

JelteF:
@tyranron feel free to sign off on after you are happy. I took a quick look and it looks reasonable to me.

impl/src/as/mod.rs
```diff
@@ -0,0 +1,371 @@
...
+                        #[automatically_derived]
+                        impl #impl_generics #trait_path for #input_type #ty_generics #where_clause {
+                            #[inline]
+                            fn #method_ident(& #mutability self) -> #return_type {
+                                #body
+                            }
+                        }
+                    }
+                })
+            .collect()
```
tyranron, line 101: @MegaBluejay I don't really like the solution here. It somewhat inverts how and where we handle the abstractions in the code. The type representing an expansion is actually expanded in the main `expand` function, while parsing/forming those expansions is made in separate functions below. It's more natural to see the opposite: we parse and form the expansion values in the main `expand` function, while the details of parsing and how exactly expansion is rendered - are abstracted behind the types and theirs APIs.

This made me to rewrite the whole thing. Please, examine the new code carefully, and consider/use this decomposition patter for future contributions.

tyranron (changes requested):
@MegaBluejay I've heavily refactored your implementation. Consider the changes.

Regarding the implementation itself, it's good enough for merging already. However, the tests are still lacking features coverage very much, which I requested in previous review. No tests for `#[as_ref]` and `#[as_ref(skip)]` were added at all. I've restructured the existing tests in `as_mut`, so they have some logical sequence, but I think it's not good enough yet.

Let's break it by modules:
````rust
mod single_field {
    mod tuple {
          // Clean
          // Forward
          // FieldForward
          mod generic {
              // Clean
              // Forward
              // FieldForward
          }
    }
    mod named {
          // Clean
          // Forward
          // FieldForward
          mod generic {
              // Clean
              // Forward
              // FieldForward
          }
    }
}
mod multi_field {
    mod tuple {
          // Clean
          // Forward
          // Field (multiple impls)
          // Ignored (multiple impls)
          mod generic {
              // Clean
              // Forward
              // Field (multiple impls)
              // Ignored (multiple impls)
          }
    }
    mod named {
          // Clean
          // Forward
          // Field (multiple impls)
          // Ignored (multiple impls)
          mod generic {
              // Clean
              // Forward
              // Field (multiple impls)
              // Ignored (multiple impls)
          }
    }
}
```

tests/as_mut.rs
```diff
@@ -1,117 +1,527 @@
...
+        }
 
-#[test]
-fn single_field_generic_struct() {
-    let mut item = SingleFieldGenericStruct { first: "test" };
+        #[derive(AsMut)]
+        struct Skip(String, i32, #[as_mut(skip)] f64);
 
-    assert!(ptr::eq(&mut item.first, item.as_mut()));
+        impl AsMut<f64> for Skip {
```
tyranron, line 267: @MegaBluejay it's a good practice to place here a comment why this impl is here, so the people don't wonder what the heck is that, and don't ask redundant questions.

tyranron (approved):
@MegaBluejay good! 👍

After merging this PR, please do as subsequent ones:

1. > Since `AsRef` and `AsMut` are effectively the same regarding implementation, should we unite them under a single `as_ref` Cargo feature as we do for `Display`-like derives?
    >
    > I think it's quite reasonable. May be done as a separate PR, though, after merging this one.

2. > After merging this PR, please open a new one which refactors `From`/`Into`/`Debug`/`Display` to reuse `utils::skip` and `utils::forward` where applicable.

PR #291, opened 2023-08-11
Add field-specific functionality to Into

### LINKED ISSUE #287: `derive(Into)` lost the ability to convert into particular fields
## Problem description

In our codebase we have the following code:
```rust
#[derive(AsRef, Clone, Debug, Default, Eq, Into)]
pub struct Data<Id> {
    id: OnceCell<Id>,
    #[as_ref]
    #[into]
    raw: String,
}
```
It worked totally fine on `0.99.*` `derive_more`  versions. But after upgrading to `1.0.0-beta.3` it stopped to compile, complaining about `#[into]` attribute being incorrect:
```
error: expected attribute arguments in parentheses: #[into(...)]
   --> whatever/paths/src/foo.rs:118:7
    |
118 |     #[into]
    |       ^^^^
```

Even if we change it as:
```rust
#[derive(AsRef, Clone, Debug, Default, Eq, Into)]
pub struct Data<Id> {
    id: OnceCell<Id>,
    #[as_ref]
    #[into(String)]
    raw: String,
}
```
It still doesn't work:
```
error: expected `skip`, found: `String`
   --> whatever/paths/src/foo.rs:118:7
    |
118 |     #[into(String)]
    |  
```

Seems like this ability was totally overlooked in #248 reimplementation and was never mentioned in documentation.

## Proposed solution

Support `#[into]` and `#[into(<types>)]` attributes on field-level for `derive(Into)` macro. Mention it in docs, and cover with tests.

### Design

Some notes about how it should interact with existing  `derive(Into)`  features:
```rust
#[derive(Into)]
struct Foo {
   a: u8,
   b: String,
}
// Generated impls:
// - impl From<Foo> for (u8, String)


#[derive(Into)]
struct Foo {
   a: u8,
   #[into]
   b: String,
}
// Generated impls:
// - impl From<Foo> for String


#[derive(Into)]
struct Foo {
   a: u8,
   #[into(Box<str>)]
   b: String,
}
// Generated impls:
// - impl From<Foo> for Box<str>


#[derive(Into)]
struct Foo {
   #[into(owned(u64), ref)]
   a: u8,
   b: String,
}
// Generated impls:
// - impl From<Foo> for u64
// - impl From<&Foo> for &u8


#[derive(Into)]
#[into(owned, ref((u8, str)), ref_mut)]
struct Foo {
   #[into(owned(u64), ref)]
   a: u8,
   b: String,
   #[into(skip)]
   _c: PhantomData<Whatever>, 
}
// Generated impls:
// - impl From<Foo> for (u8, String)
// - impl From<&Foo> for (&u8, &str)
// - impl From<&mut Foo> for (&mut u8, &mut String)
// - impl From<Foo> for u64
// - impl From<&Foo> for &u8
```
And, maybe, it has sense to allow this case:
```rust
#[derive(Into)]
#[into]
struct Foo {
   #[into(owned(u64), ref)]
   a: u8,
   #[into]
   b: String,
   #[into(skip)]
   _c: PhantomData<Whatever>, 
}
// Generated impls:
// - impl From<Foo> for (u8, String)
// - impl From<Foo> for u64
// - impl From<&Foo> for &u8
// - impl From<Foo> for String
```
So, we have the parity of how `#[into(<types>)]` attribute works on top-level and field-level.

MegaBluejay:
Resolves #287 

## Synopsis

Allow using the same attribute arguments on fields as on the whole struct, generate `impl`s for each

## Solution

Extract the common arguments of struct and field attributes into a seperate parser

See details on proposed semantics of using both together in the issue

## Checklist

[l:1]: /CHANGELOG.md


tests/into.rs
```diff
@@ -1043,3 +1043,366 @@ mod multi_field {
...
+            fn from(value: Tuple) -> Self {
+                (value.0, value.1, value.2)
+            }
+        }
+
+        impl From<Tuple> for (i32, f64) {
+            fn from(value: Tuple) -> Self {
+                (value.0, value.2)
+            }
+        }
```
JelteF, line 1066: Why are these impls here?

MegaBluejay: The idea was to test that no extra impls get generated.

For instance, no impl should be generated using all the fields if there's at least one field attribute, and if there are types specified in an attribute the field's type shouldn't  get one either.

tyranron (changes requested):
@MegaBluejay please, read the https://github.com/JelteF/derive_more/issues/287#issuecomment-1676285873, where the design was finalized, and consider it.

> A case currently not supported is skipping fields for which other arguments are also present, e.g.
>
> ```rust
> #[derive(Into)]
> #[into]
> struct Foo {
>     #[into(skip, String)]
>     a: String,
>     b: i32
>     c: f64,
> }
> ```
>
> to generate `Foo -> (i32, f64)` and `Foo -> String` impls.
>
> This isn't mentioned in the issue, and I'm not sure how useful it would be

This should be definitely supported as separate attributes:
```rust
#[derive(Into)]
#[into]
struct Foo {
    #[into(String)]
    #[into(skip)]
    a: String,
    b: i32
    c: f64,
}
```

tyranron (changes requested):
@MegaBluejay please, consider to reuse #303 changes in this PR.

tyranron (approved):
@MegaBluejay thanks! Overall: good job! 🍻

I've simplified your solution a bit, as introducing generic `Alt` and `Pair` structural types encoding just for removing `Option` merging boilerplate feels a little bit too much. Adding `ParseMultiple::merge_opt_attrs()` should do the job as good, while keeping less amount of code overall.

@JelteF since this PR introduces new API, please take a look.

Also, after merging this PR, would you be so kind to release `1.0.0-beta.4` version?

tyranron:
ping @JelteF

JelteF (approved):

JelteF:
The new API looks good to me

PR #293, opened 2023-08-14
Add support for source forwarding in Error derive

### LINKED ISSUE #428: Support for `#[error(forward)]`
Could you consider adding support for `#[error(forward)]` which provide the same functionality as `#[error(transparent)]` in the [thiserror](https://docs.rs/thiserror/latest/thiserror/) crate? (Use cases are explained there.)

In short, this would make `fn source()` return `self.source.as_dyn_error().source()` instead of `Some(self.source.as_dyn_error())`.

I can use thiserror, but derive_more is a crate I use everywhere, and it would be nice to have this functionality without depending on thiserror.

### LINKED ISSUE #511: `#[derive(Error)]` fails on enum variant with a lifetime.
## Reproduction code

```rust
#![allow(unused)]

use derive_more::{Display, Error};

pub struct Owned {
    #[error(not(source))]
    message: String,
}

pub struct Borrowed<'a> {
    #[error(not(source))]
    message: &'a str,
}

pub enum MyError<'a> {
    Owned(Owned),
    Borrowed(Borrowed<'a>),
}
```

## Compilation error

```log
error: lifetime may not live long enough
  --> src/tmp_err_report.rs:17:26
   |
17 | #[derive(Debug, Display, Error)]
   |                          ^^^^^ returning this value requires that `'a` must outlive `'static`
18 | pub enum MyError<'a> {
   |                  -- lifetime `'a` defined here
   |
   = note: this error originates in the derive macro `Error` (in Nightly builds, run with -Z macro-backtrace for more info)
```

## Current workaround

Implement `Error` manually:

```rust
#[derive(Debug, Display)]
pub enum MyError<'a> {
    Owned(Owned),
    Borrowed(Borrowed<'a>),
}

impl<'a> core::error::Error for MyError<'a> {
    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
        match self {
            MyError::Owned(error) => error.source(),
            MyError::Borrowed(error) => error.source(),
        }
    }
}
```

MegaBluejay:
Resolves #428   
Related to #110, #200 

## Synopsis

Adds support for `#[error(forward)]` attribute, which forwards the `.source()` implementation to a field.

## Solution

Use existing forward fields in `State` for parsing the attributes, and keep the source-inferring logic the same as without forwarding.

## Backtrace forwarding

The current behaviour is that if the field annotated/inferred to be the source is annotated with `#[error(backtrace)]`, a forwarded `provide()` is generated.

This was in the example usage, but not documented in the list above, so I added it.

## Checklist

  - [x] Document new behaviour
  - [x] Add an explanation of advised usage as [previously suggested](https://github.com/JelteF/derive_more/pull/110#issuecomment-588113801)?
- [x] Add errors on usage of `#[error(forward)]` when no source field was specified/inferred

[l:1]: /CHANGELOG.md


MegaBluejay:
There were some updates to the [`error_generic_member_access`](https://github.com/rust-lang/rust/issues/99301) API recently, which is why nightly tests are failing. I'll open a separate PR to fix this.

MegaBluejay:
@tyranron I implemented this with the existing parsing since it was an easy change.

Would it be preferable to refactor the derive first, similarly to #286?

tyranron:
@MegaBluejay no, let's postpone refactoring here as a minor priority task.

impl/doc/error.md
```diff
@@ -32,6 +32,11 @@ often [`From` as well](crate::From).
    called `Backtrace`. Then it would return that field as the `backtrace`.
 3. One of the fields is annotated with `#[error(backtrace)]`. Then it would
    return that field as the `backtrace`.
+4. The source field is annotated with `#[error(backtrace)]`. Then it would
```
JelteF, line 35: copy paste error?
```suggestion
4. The source field is annotated with `#[error(forward)]`. Then it would
```

JelteF:
I'm having a hard time understanding the usecase for such forwarding. Could you share a concrete example for when this is useful?

impl/doc/error.md
MegaBluejay: No that's correct, `provide` is forwarded only when the source is annotated with `#[error(backtrace)]`.

That behaviour is unchanged in this PR.

The reasoning for keeping a separate attribute is to make source forwarding work on stable, since there's no good way of generating feature-gated code.

I also think this is somewhat unintuitive, but I haven't come up with a better idea.

Perhaps it could make sense to disallow `#[error(backtrace)]` on the source field if there's no `#[error(forward)]`?

tyranron:
@MegaBluejay yes, I start thinking that maybe `transparent` is better here than `forward`. Let's postpone it for a little while... I'll write up some design thoughts during the following week.

JonahPlusPlus:
Example use case:
I want to have a composite error type, and forwarding the source would simplify deriving Display and Error.

e.g.
```rust
#[display("`{rust_name}` should work but {kind}")]
struct Error {
  rust_name: &'static str,
  #[error(forward)]
  kind: ErrorKind
}

enum ErrorKind {
  #[display("foo was encountered")]
  Foo,
  #[display("bar errored")]
  Bar {
    #[error(source)]
    external: ExternalError
  }
}
```

We have data common to all variants, so moving it into a struct makes sense. But we also have source information only relevant to specific variants, so we need some way to use the source in the enum and we don't want to treat the `ErrorKind` as it's own source, since the enum doesn't represent an "abstraction barrier" by itself. So forwarding the error makes the most sense here.

Note that this pattern isn't available in thiserror, since `#[error(transparent)]` also delegates the `Display` impl.

nerditation:
> I start thinking that maybe `transparent` is better here than `forward`

what's the current status on this matter? it seems to be stalled for some time?

I agree `#[error(transparent)]` would be useful, and it is what other libraries like `thiserror` and `snafu` already have.

on the other hand, I would also like to have an orthogonal attribute to forward the `source()` method call, e.g. something like `#[error(source(forward))]`, and maybe `#[error(backtrace(forward))]` too, for completeness. (well, this is not really orthogonal to, but probably mutual exclusive with `transparent`).

they serve different purpose and use cases, one example is shown by @JonahPlusPlus, where an error type shares some common information, while each variant has a different source, so it makes sense to use a `struct Error` for the shared fields, and use an `enum ErrorKind` to distinguish the different error cases and their sources.

in a manual implementation, I can match `self.kind` directly to extract the source, something like:

```rust
struct Error {
    // shared by all cases
    object_id: i32
    //...
    // different per error case
    kind: ErrorKind,
}
enum ErrorKind {
    OutOfBound, // no source error
    Read(io::Error), // io error
    ParseX(num:ParseIntError), // parsing error for `X`
    ParseY(num:ParseIntError), // for `Y`
    //...etc
}
impl StdError for Error {
    fn source(&self) -> Option<&dyn StdError + 'static> {
        match self.kind {
            //...
        }
    }
}
```

but for a procedural macro, a derive on `struct Error` cannot see the definition of `enum ErrorKind`, so the pattern matching must be derived on `ErrorKind`, but it is logically not a layer in the error stack, so I don't want to use it as the "source" of my `struct Error` type, and an attribute to forward just the `source()` method call would be great in this situation.

this design is also flexible, e.g. the `ErrorKind` might just holds the sources initially, but later each error case might add specific fields to the variant, then the `ErrorKind` itself can logically be the source of `Error`, so we can remove the `forward` annotation, e.g. `#[error(source(forward))]` to `#[error(source)]`, and it still works. on the other hand, we could remove the common fields in the struct and (selectively) move them to some variants, in which case the `struct Error` becomes a real *transparent* wrapper of `ErrorKind`, and that still works too.

we may bikeshed the name, `forward`, `delegate`, `passthrough`, whatever, but we need to make the decision to support it first.

PR #294, opened 2023-08-15
Fix generated `Error::provide` and tests to reflect changes in API

MegaBluejay:
## Synopsis

Update derive and tests to work on the latest nightly with the changes in [this PR](https://github.com/rust-lang/rust/pull/113464).

## Solution

Use `error::{Request, request_ref, request_value}` instead of `any::{Demand, request_ref, request_value}`

Currently `request_value` is not re-exported from `core::error` in `std::error`, which means backtrace tests require `#![feature(error_in_core)]`

Presumably this is a bug, and will be fixed later

## Checklist

[l:1]: /CHANGELOG.md


tyranron (approved):
@MegaBluejay nice job! Thanks 🍻

PR #295, opened 2023-08-18
Unite `as_ref` and `as_mut` features

MegaBluejay:
First suggested [here](https://github.com/JelteF/derive_more/pull/286#issuecomment-1678901955)

## Synopsis

`AsRef` and `AsMut` derives are effectively the same regarding implementation, so using separate features for them doesn't make sense.

## Solution

Use the `as_ref` feature for both derives

## Checklist

[l:1]: /CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,9 +36,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - The `Into` derive now uses `#[into(<types>)]` instead of `#[into(types(<types>))]`
   and ignores field type itself.
 - Importing a derive macro now also import its corresponding trait.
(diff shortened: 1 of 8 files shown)


MegaBluejay:
We could keep the `as_mut` feature around as a synonym, making this a non-breaking change.
As far as I understand removing it from the public interface is part of the goal here, so I didn't.

tyranron:
@MegaBluejay we're going to `1.0.0`, so breaking changes are okay now, not later. I think that keeping `as_mut` as synonym will make even more confusion to library users. It's okay to have just `as_ref`, just `deref`, and so on...

tyranron (dismissed):

tyranron (approved):

PR #296, opened 2023-08-18
Use the new parsing utils

MegaBluejay:
First suggested [here](https://github.com/JelteF/derive_more/pull/286#discussion_r1297046342)

## Synopsis

There's some repetition around parsing the same attributes in different derives

## Solution

Refactor `Into`, `From`, and `Debug` derives to use the `skip` and `forward` attribute parsing utils

## Checklist

[l:1]: /CHANGELOG.md

### DIFF
--- a/impl/src/utils.rs
+++ b/impl/src/utils.rs
@@ -25,7 +25,7 @@ pub(crate) use self::either::Either;
 
 #[cfg(any(feature = "from", feature = "into"))]
 pub(crate) use self::fields_ext::FieldsExt;
-#[cfg(any(feature = "as_ref", feature = "as_mut"))]
+#[cfg(any(feature = "as_ref", feature = "as_mut", feature = "from"))]
 pub(crate) use self::spanning::Spanning;
 
 #[derive(Clone, Copy, Default)]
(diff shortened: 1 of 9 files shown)


impl/src/utils.rs
```diff
@@ -1385,6 +1397,31 @@ pub(crate) mod skip {
     /// ```
     pub(crate) struct Attribute(&'static str);
 
+    impl Attribute {
```
tyranron: @MegaBluejay why as a separate `impl Attribute` block? There is one already down bellow.

impl/src/utils.rs
```diff
@@ -1385,6 +1397,31 @@ pub(crate) mod skip {
...
+        pub(crate) fn parse_attrs(
```
tyranron, line 1422: @MegaBluejay why without docs, again?

impl/src/utils.rs
```diff
@@ -1370,13 +1375,20 @@ pub(crate) mod forward {
     }
 }
 
-#[cfg(feature = "as_ref")]
+#[cfg(any(
+    feature = "as_ref",
+    feature = "from",
+    feature = "into",
+    feature = "debug",
```
tyranron: @MegaBluejay always preserve the alphabetic order, by default.

tyranron (approved):

PR #300, opened 2023-08-31
Add `TryFrom` to convert repr to enum

### LINKED ISSUE #146: Derive `TryFrom<T>` for enums marked `#[repr(T)]`
C-like enums with an explicit `#[repr(<int type>)]` attribute can be cast to that integer type, but there's no easy way to convert back:

```rust
#[repr(i16)]
enum MyEnum {
    Foo = 1,
    Bar = 2,
}

let val = MyEnum::Bar as i16;
let my_enum: MyEnum = // convert `val` back to `MyEnum`
```

This derive exists in [`enum-utils`](https://docs.rs/enum-utils/0.1.2/enum_utils/derive.TryFromRepr.html) but it would be nice not to have to reach for multiple crates.

ModProg:
Resolves #146

## Synopsis

Conversion from the repr integer to an enum.

## Solution

A `TryFrom` derive macro implementing `TryFrom<integertype> for Enum`.

## Checklist

    - [x] basic tests
    - [x] compile fail

[l:1]: /CHANGELOG.md

### DIFF
--- a/impl/doc/try_from.md
+++ b/impl/doc/try_from.md
@@ -12,6 +12,7 @@ Only field-less variants can be constructed from their variant, therefor the `Tr
 # mod discriminant_on_non_unit_enum {
 # use derive_more::TryFrom;
 #[derive(TryFrom, Debug, PartialEq)]
+#[try_from(repr)]
 #[repr(u32)]
 enum Enum {
     Implicit,
--- a/impl/doc/try_from.md
+++ b/impl/doc/try_from.md
@@ -1,0 +1,37 @@
+# What `#[derive(TryFrom)]` generates
+
+This derive allows you to convert enum discriminants into their corresponding variants.
+By default a `TryFrom<isize>` is generated, matching the [type of the discriminant](https://doc.rust-lang.org/reference/items/enumerations.html#discriminants).
+The type can be changed with a `#[repr(u/i*)]` attribute, e.g., `#[repr(u8)]` or `#[repr(i32)]`.
+Only field-less variants can be constructed from their variant, therefor the `TryFrom` implementation will return an error for a discriminant representing a variant with fields.
+
+## Example usage
+
+```rust
+# #[rustversion::since(1.66)]
+# mod discriminant_on_non_unit_enum {
+# use derive_more::TryFrom;
+#[derive(TryFrom, Debug, PartialEq)]
+#[repr(u32)]
+enum Enum {
+    Implicit,
+    Explicit = 5,
+    Field(usize),
+    Empty{},
+}
+
+# #[rustversion::since(1.66)]
+# pub fn test(){
+assert_eq!(Enum::Implicit, Enum::try_from(0).unwrap());
+assert_eq!(Enum::Explicit, Enum::try_from(5).unwrap());
+assert_eq!(Enum::Empty{}, Enum::try_from(7).unwrap());
+
+// variants with fields are not supported
+assert!(Enum::try_from(6).is_err());
+# }
+# }
+# // We need to use a `function` declaration, because we cannot put `rustversion` on a statement.
+# #[rustversion::since(1.66)] use discriminant_on_non_unit_enum::test;
+# #[rustversion::before(1.66)] fn test() {}
+# test();
+```
--- a/impl/doc/try_from.md
+++ b/impl/doc/try_from.md
@@ -8,8 +8,6 @@ Only field-less variants can be constructed from their variant, therefor the `Tr
 ## Example usage
 
 ```rust
-# #[rustversion::since(1.66)]
-# mod discriminant_on_non_unit_enum {
 # use derive_more::TryFrom;
 #[derive(TryFrom, Debug, PartialEq)]
 #[try_from(repr)]
@@ -21,18 +19,10 @@ enum Enum {
     Empty{},
 }
 
-# #[rustversion::since(1.66)]
-# pub fn test(){
 assert_eq!(Enum::Implicit, Enum::try_from(0).unwrap());
 assert_eq!(Enum::Explicit, Enum::try_from(5).unwrap());
 assert_eq!(Enum::Empty{}, Enum::try_from(7).unwrap());
 
-// variants with fields are not supported
+// Variants with fields are not supported, as the value for their fields would be undefined.
 assert!(Enum::try_from(6).is_err());
-# }
-# }
-# // We need to use a `function` declaration, because we cannot put `rustversion` on a statement.
-# #[rustversion::since(1.66)] use discriminant_on_non_unit_enum::test;
-# #[rustversion::before(1.66)] fn test() {}
-# test();
 ```
--- a/impl/doc/try_from.md
+++ b/impl/doc/try_from.md
@@ -13,15 +13,15 @@ Only field-less variants can be constructed from their variant, therefor the `Tr
 #[try_from(repr)]
 #[repr(u32)]
 enum Enum {
-    Implicit,
-    Explicit = 5,
-    Field(usize),
-    Empty{},
+    ImplicitZero,
+    ExplicitFive = 5,
+    FieldSix(usize),
+    EmptySeven{},
 }
 
-assert_eq!(Enum::Implicit, Enum::try_from(0).unwrap());
-assert_eq!(Enum::Explicit, Enum::try_from(5).unwrap());
-assert_eq!(Enum::Empty{}, Enum::try_from(7).unwrap());
+assert_eq!(Enum::ImplicitZero, Enum::try_from(0).unwrap());
+assert_eq!(Enum::ExplicitFive, Enum::try_from(5).unwrap());
+assert_eq!(Enum::EmptySeven{}, Enum::try_from(7).unwrap());
 
 // Variants with fields are not supported, as the value for their fields would be undefined.
 assert!(Enum::try_from(6).is_err());
--- a/impl/doc/try_from.md
+++ b/impl/doc/try_from.md
@@ -1,11 +1,15 @@
 # What `#[derive(TryFrom)]` generates
 
-This derive allows you to convert enum discriminants into their corresponding variants.
-By default a `TryFrom<isize>` is generated, matching the [type of the discriminant](https://doc.rust-lang.org/reference/items/enumerations.html#discriminants).
+Derive `TryFrom` allows you to convert enum discriminants into their corresponding variants.
+
+
+
+
+## Enums
+
+By default, a `TryFrom<isize>` is generated, matching the [type of the discriminant](https://doc.rust-lang.org/reference/items/enumerations.html#discriminants).
 The type can be changed with a `#[repr(u/i*)]` attribute, e.g., `#[repr(u8)]` or `#[repr(i32)]`.
 Only field-less variants can be constructed from their variant, therefor the `TryFrom` implementation will return an error for a discriminant representing a variant with fields.
-
-## Example usage
 
 ```rust
 # use derive_more::TryFrom;
--- a/impl/src/try_from.rs
+++ b/impl/src/try_from.rs
@@ -165,14 +165,14 @@ impl ToTokens for Expansion {
                  ::core::convert::TryFrom<#repr #ty_generics> for #ident
                  #where_clause
             {
-                type Error = ::derive_more::TryFromError<#repr>;
+                type Error = ::derive_more::TryFromReprError<#repr>;
 
                 #[inline]
                 fn try_from(value: #repr) -> ::core::result::Result<Self, Self::Error> {
                     #(#[allow(non_upper_case_globals)] const #consts: #repr = #discriminants;)*
                     match value {
                         #(#consts => ::core::result::Result::Ok(#ident::#variants),)*
-                        _ => ::core::result::Result::Err(::derive_more::TryFromError::new(value)),
+                        _ => ::core::result::Result::Err(::derive_more::TryFromReprError::new(value)),
                     }
                 }
             }
--- a/impl/src/try_from.rs
+++ b/impl/src/try_from.rs
@@ -1,7 +1,7 @@
 //! Implementation of a [`TryFrom`] derive macro.
 
 use proc_macro2::{Literal, Span, TokenStream};
-use quote::{format_ident, quote, ToTokens as _};
+use quote::{format_ident, quote, ToTokens};
 use syn::spanned::Spanned as _;
 
 /// Expands a [`TryFrom`] derive macro.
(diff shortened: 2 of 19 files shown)


impl/doc/try_from.md
```diff
@@ -0,0 +1,36 @@
...
+assert_eq!(Enum::Empty{}, Enum::try_from(7).unwrap());
+
+// variants with fields are not supported
+assert!(Enum::try_from(6).is_err());
+# }
+# }
+# // We need to use a `function` declaration, because we cannot put `rustversion` on a statement.
+# #[rustversion::since(1.66)] use discriminant_on_non_unit_enum::test;
+# #[rustversion::before(1.66)] fn test() {}
+# test();
```
ModProg, line 35: @JelteF I'm really not sure how to do this better, the problem is, that 1.65 doesn't allow discriminants on enums with fields.

impl/src/try_from.rs
```diff
@@ -0,0 +1,149 @@
...
+                    let repr: Ident = attr.parse_args()?;
+                    match repr.to_string().as_str() {
+                        "u8" | "u16" | "u32" | "u64" | "u128" | "usize" | "i8"
+                        | "i16" | "i32" | "i64" | "i128" | "isize" => Ok(Some(repr)),
+                        _ => Err(syn::Error::new(
+                            repr.span(),
+                            "only integer discriminants (u*/i*) are supported",
+                        )),
+                    }
+                }
```
ModProg, line 67: There is one todo here, we should also support:
```rs
#[repr(C, u8)]

#[repr(u8, align(4))]

#[repr(align(4))]
#[repr(u8)]
```
etc.

impl/src/try_from.rs
```diff
@@ -0,0 +1,149 @@
...
+            data.struct_token.span(),
+            "`TryFrom` cannot be derived for structs",
+        )),
+        syn::Data::Enum(data) => {
+            let Some(repr) = ReprAttribute::parse_attrs(&input.attrs)? else {
+                return Err(syn::Error::new(
+                    Span::call_site(),
+                    "`TryFrom` needs repr specified `#[repr(u*/i*)]`",
+                ));
+            };
```
ModProg, line 20: This is probably overly conservative, we should be able to assume that it is an `isize` if nothing was specified:

> Under the [default representation](https://doc.rust-lang.org/reference/type-layout.html#the-default-representation), the discriminant is interpreted as an isize value. However, the compiler is allowed to use a smaller type (or another means of distinguishing variants) in its actual memory layout.

https://doc.rust-lang.org/reference/items/enumerations.html#discriminants

src/convert.rs
```diff
@@ -45,3 +45,35 @@ impl<T> fmt::Display for TryIntoError<T> {
...
+// `T` should only be an integer type and therefor display
+impl<T: fmt::Display> fmt::Display for TryFromError<T> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "`{}` does not respond to a unit variant", self.input)
+    }
+}
+
+#[cfg(feature = "std")]
+// `T` should only be an integer type and therefor display and debug
+impl<T: fmt::Debug + fmt::Display> std::error::Error for TryFromError<T> {}
```
ModProg, line 79: Not sure what good error message would be here.

ModProg:
We could also support some `#[try_from(<type>)]` for types whose definition is a superset to the repr type.

impl/src/try_from.rs
ModProg: Got another confirmation on the `isize` for the expression especially: https://rust-lang.zulipchat.com/#narrow/stream/182449-t-compiler.2Fhelp/topic/enum.20discriminant.20size/near/388599445

impl/doc/try_from.md
tyranron: @ModProg @JelteF I think it's worth bumping MSRV up to 1.66.

ModProg: On the one hand it's only documentation, on the other hand, it's only one version.

tests/compile_fail/try_from/invalid_repr.rs
```diff
@@ -0,0 +1,7 @@
+#[derive(derive_more::TryFrom)]
+#[repr(a + b)]
+enum Enum {
```
tyranron, line 3: @ModProg it would also be nice to cover the case with _valid_ `repr`, but not meaningful, like `repr(Rust)` or `repr(C)`.

I think, we should control the exact list of types we accept, inside the macro: `i8`, `u8`, `i16`, `u16`, `i32`, `u32`, `i64`, `u64`, `i128`, `u128`, `isize`, `usize`. And disregard anything else.

src/convert.rs
```diff
@@ -45,3 +45,35 @@ impl<T> fmt::Display for TryIntoError<T> {
...
+    #[doc(hidden)]
+    #[must_use]
+    #[inline]
+    pub const fn new(input: T) -> Self {
+        Self { input }
+    }
+}
+
+// `T` should only be an integer type and therefor display
+impl<T: fmt::Display> fmt::Display for TryFromError<T> {
```
tyranron, line 71: @ModProg that's too limiting bound. Usually, most of types don't impl `Display`, and thus we make the error here non-`Display`able for the majority of cases. Better remake the error message the way we lift this trait bound. I think adding a type name as `&'static str` to the error would do the trick, as it does for the `TryIntoError` above.

impl/doc/try_from.md
```diff
@@ -0,0 +1,37 @@
...
+}
+
+# #[rustversion::since(1.66)]
+# pub fn test(){
+assert_eq!(Enum::Implicit, Enum::try_from(0).unwrap());
+assert_eq!(Enum::Explicit, Enum::try_from(5).unwrap());
+assert_eq!(Enum::Empty{}, Enum::try_from(7).unwrap());
+
+// variants with fields are not supported
+assert!(Enum::try_from(6).is_err());
```
tyranron, line 30: @ModProg this very rare and uncommon situation, so I think library users deserve some reasoning and explanation why this is not supported exactly, because it's quite unclear, imho.

impl/src/try_from.rs
```diff
@@ -0,0 +1,141 @@
...
+    fn parse_attrs(attrs: impl AsRef<[syn::Attribute]>) -> syn::Result<Self> {
+        attrs
+            .as_ref()
+            .iter()
+            .filter(|attr| attr.path().is_ident("repr"))
+            .try_fold(None, |mut repr, attr| {
+                attr.parse_nested_meta(|meta| {
+                    if let Some(ident) = meta.path.get_ident() {
+                        if let "u8" | "u16" | "u32" | "u64" | "u128" | "usize" | "i8"
+                        | "i16" | "i32" | "i64" | "i128" | "isize" =
```
tyranron, line 46: > I think, we should control the exact list of types we accept, inside the macro: `i8`, `u8`, `i16`, `u16`, `i32`, `u32`, `i64`, `u64`, `i128`, `u128`, `isize`, `usize`. And disregard anything else.

@ModProg you do so, already. Good 👍

So we need only a test covering that.

impl/src/try_from.rs
```diff
@@ -0,0 +1,141 @@
...
+                        }
+                    }
+                    // ignore all other attributes that could have a body e.g. `align`
+                    _ = meta.input.parse::<proc_macro2::Group>();
+                    Ok(())
+                })
+                .map(|_| repr)
+            })
+            // Default discriminant is interpreted as `isize` (https://doc.rust-lang.org/reference/items/enumerations.html#discriminants)
+            .map(|repr| repr.unwrap_or_else(|| Ident::new("isize", Span::call_site())))
```
tyranron, line 60: @ModProg let's mention this in docs too. Quite unclear and subtle moment.

impl/src/try_from.rs
```diff
@@ -0,0 +1,141 @@
...
+
+/// Expansion of a macro for generating [`TryFrom`] implementation of an enum
+struct Expansion<'a> {
+    /// Enum `#[repr(u/i*)]`
+    repr: ReprAttribute,
+    /// Enum [`Ident`].
+    ident: &'a Ident,
+
+    /// Variant [`Ident`] in case of enum expansion.
+    variants: Vec<&'a syn::Variant>,
```
tyranron, line 73: @ModProg I think it's not worth here to bother with lifetimes and complicating signatures. Just `Clone`/`Copy` is enough. It won't give any real performance benefit here, so the library users can feel it.

impl/doc/try_from.md
JelteF: I would say let's just bump it to 1.72, by the time we release 1.0 I think that will have been out long enough.

impl/src/try_from.rs
```diff
@@ -0,0 +1,141 @@
...
+
+    /// Struct or enum [`syn::Generics`].
+    generics: &'a syn::Generics,
+}
+
+impl<'a> Expansion<'a> {
+    /// Expands [`TryFrom`] implementations for a struct.
+    fn expand(&self) -> syn::Result<TokenStream> {
```
tyranron, line 81: @ModProg it's better and more natural to write expansion as `ToTokens` impl. This also prevents the expansion to be fallible. Anything that can error should be checked before performing the expansion.

tyranron (changes requested):
@ModProg good job! Thanks 👍
Overall design is solid and thoughtful. The implementation is also good, despite few bikesheddings described below.

> We could also support some `#[try_from(<type>)]` for types whose definition is a superset to the repr type.

Definitely not as a part of this PR. Let's sketch and discuss the design of it in #146.

src/convert.rs
ModProg: Maybe it would make sense to have an explicit `TryFromRepr` error, then? Because all types available as repr do in fact implement `Display`.

impl/doc/try_from.md
ModProg: Maybe:
> Variants with fields are not supported, as the value for their fields would be undefined.

impl/src/try_from.rs
ModProg: https://github.com/JelteF/derive_more/blob/819cb3d2b98e08c941b4747b3df70d4d15aceca2/impl/doc/try_from.md?plain=1#L4-L5

impl/src/try_from.rs
ModProg: Probably true, I took this and the fallible `expand` from the `From` implementation

tyranron:
@ModProg @JelteF yet another thing that i think of: we will definitely want to support this case in the future:
```rust
#[derive(TryFrom)]
enum Foo {
    #[try_from(u128)] // generates `impl TryFrom<u128> for Foo`
    Bar(u8),
}
```
While at the same time, in other situations we would like to have something like this:
```
#[derive(TryFrom)]
#[try_from(u128)] // generates `impl TryFrom<u128> for Foo`
#[repr(u8)]
enum Foo {
    Bar,
}
```
So, we need a way to differentiate a `repr` conversions from the ones of the inner types clearly.

And this question touches not only `TryFrom`, because we also will want to have something like this for `Into`:
```rust
#[derive(Into)] // generates `impl From<Foo> for u8`
#[repr(u8)]
enum Foo {
    Bar = 35,
}
```
And `TryInto`:
```rust
#[derive(TryInto)]
#[try_into(u8)] // generates `impl TryFrom<Foo> for u8`
#[repr(u128)] 
enum Foo {
    Bar = 35,
}
```

To disambiguate them clearly, I propose to use `repr` sub-attribute:
```rust
#[derive(Into, TryFrom)]
#[into(repr)] // generates `impl From<Foo> for u8`
#[try_from(repr)] // generates `impl TryFrom<u8> for Foo`
#[repr(u8)]
enum Foo1 {
    Bar,
}

#[derive(TryFrom, TryInto)]
#[try_from(repr(u128))] // generates `impl TryFrom<u128> for Foo`
#[try_into(repr(u8))] // generates `impl TryFrom<Foo> for u8`
#[repr(u16)]
enum Foo2 {
    Bar,
}
```

Does it make sense?

ModProg:
@tyranron, looks sensible. I'll implement support for the repr syntax for now, and will error on it not specified.

ModProg:
@tyranron how would you expect `repr(<types>)` to work?

```
impl TryFrom<$type> for $enum {
    fn try_from(value: $type) -> Result<Self> {
        let value: $actual_repr = value.try_into().map_err(|e| TryFromError::message(e.to_string())?;
        match value {
            $actual_implementation
        }
    }
}
```

tyranron:
@ModProg as a regular forwarding implementation, like you've pointed, yes. But as I've [mentioned before](https://github.com/JelteF/derive_more/pull/300#pullrequestreview-1612916424), let's not making it as a part of this PR. Better implement as a separate one after merging this.

impl/src/try_from.rs
tyranron: @ModProg seems to be enough

src/convert.rs
tyranron: @ModProg yes, makes sense.

impl/doc/try_from.md
```diff
@@ -0,0 +1,28 @@
...
+```rust
+# use derive_more::TryFrom;
+#[derive(TryFrom, Debug, PartialEq)]
+#[try_from(repr)]
+#[repr(u32)]
+enum Enum {
+    Implicit,
+    Explicit = 5,
+    Field(usize),
+    Empty{},
```
JelteF, line 19: I think the example is good, but hard to understand at a glance.
```suggestion
    ImplicitZero,
    ExplicitFive = 5,
    FieldSix(usize),
    EmptySeven{},
```

src/convert.rs
```diff
@@ -45,3 +45,36 @@ impl<T> fmt::Display for TryIntoError<T> {
...
+impl<T> TryFromReprError<T> {
+    #[doc(hidden)]
+    #[must_use]
+    #[inline]
+    pub const fn new(input: T) -> Self {
+        Self { input }
+    }
+}
+
+// `T` should only be an integer type and therefor be debug
```
JelteF, line 71:
```suggestion
// `T` should only be an integer type and therefore be debug
```

src/convert.rs
```diff
@@ -45,3 +45,36 @@ impl<T> fmt::Display for TryIntoError<T> {
...
+impl<T: fmt::Debug> fmt::Display for TryFromReprError<T> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "`{:?}` does not respond to a unit variant", self.input)
```
JelteF, line 74:
```suggestion
        write!(f, "`{:?}` does not corespond to a unit variant", self.input)
```

JelteF (dismissed):
Looks good to me (apart from the few small comments I left). @tyranron did you have any more feedback.

impl/doc/try_from.md
```diff
@@ -0,0 +1,28 @@
...
+}
+
+assert_eq!(Enum::Implicit, Enum::try_from(0).unwrap());
+assert_eq!(Enum::Explicit, Enum::try_from(5).unwrap());
+assert_eq!(Enum::Empty{}, Enum::try_from(7).unwrap());
```
ModProg, line 24:
```suggestion
assert_eq!(Enum::ImplicitZero, Enum::try_from(0).unwrap());
assert_eq!(Enum::ExplicitFive, Enum::try_from(5).unwrap());
assert_eq!(Enum::EmptySeven{}, Enum::try_from(7).unwrap());
```

tyranron:
@JelteF yes, I was waiting for #298 to land. Give me a moment.

tyranron:
@JelteF oh, and please, change the required CI jobs in project settings to reflect MSRV 1.65.0 -> MSRV 1.72.0 migration. Otherwise, CI won't pass. Thanks!

tyranron (approved):

PR #313, opened 2023-11-11
chore: lower MSRV to 1.65

DaniPopes:
## Synopsis

The MSRV bump in https://github.com/JelteF/derive_more/pull/300 is unnecessarily high. The 1.72 minimum is only needed for tests, but the actual implementation depends on 1.65.

## Solution

Lower MSRV from 1.72 to 1.65. Only `build` instead of `test` in MSRV CI job.

## Checklist

[l:1]: /CHANGELOG.md


tyranron (dismissed):
@DaniPopes thanks for taking care of the details!

@JelteF please, update the required CI jobs in project settings, and merge.

tyranron (approved):

tyranron:
ping @JelteF

JelteF:
I think it might be nice to have a bit higher MSRV than strictly necessary, because we'll likely keep it the same for the next few years. So I'm not sure we should merge this PR

tyranron:
@JelteF treating MSRV via SemVer of the crate seems to be quite uncomfortable: either we change the major version when don't really break API, or we are committed to not use newer Rust features until the real API break happens.

Considering that MSRV is not that much of the demand for end users (people tend to use newer versions of Rust), we could clearly state (in README) and separate this in the following manner:
- Changing MSRV is treated as a __minor version__ in terms of SemVer.
- So, If you're not concerned about MSRV in your projects, you just use the default [caret requirement](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#caret-requirements): `derive_more = "1"`, or `derive_more = "1.0"`, or `derive_more = "^1.0"`.
- However, if you're concerned about MSRV in your projects, then just use the [tilde requirement](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#tilde-requirements) to omit breaks: `derive_more = "~1.0"`, or `derive_more = "~1.0.0"`

tyranron:
@JelteF ping

tyranron:
@JelteF ping

JelteF:
I think your proposal sounds good, but we should update the readme accordingly

tyranron (approved):
@JelteF I've added "MSRV policy" section to README.

I've also returned running `cargo test` in `msrv` CI job, because it's vital to check macro expansion too for MSRV. Anything that doesn't fit MSRV from user-side in tests, I've hidden under the `#[cfg(not(msrv))]`.

I'll merge this tomorrow, so you can have a look if you want to.

PR #318, opened 2023-11-17
Fix codegen fail tests for 1.74 Rust

tyranron:


tyranron:
Not required for this version of Rust.

PR #322, opened 2023-12-15
Delegation for `fmt` traits (#321)

### LINKED ISSUE #321: FR: Add `#[display(delegate)]` option for newtypes which delegates to wrapped type
Add option to allow *delegating* Display implementation to wrapped field or to implementation of another formatting trait.

----

Currently, using something akin to `#[display(fmt = "{}", _0)]` for newtypes suffers from the issues of all user-provided flags being reset.  For example:

```rust
#[derive(derive_more::Display)]
#[display(fmt = "{}", _0)]
struct Num(usize);

struct Expected(usize);

impl std::fmt::Display for Expected {
    fn fmt(&self, fmtr: &mut std::fmt::Formatter) -> std::fmt::Result {
        self.0.fmt(fmtr)
    }
}

fn main() {
    println!("got={:03} want={:03}", Num(7), Expected(7));
        // → got=7 want=007
}
```

---

The idea is to support `#[dispaly(delegate)]` annotation which would delegate call rather than use `write!` macro.  For example:

```rust
#[derive(derive_more::Display)]
#[display(delegate)]
struct Num(usize);
```

would generate:

```rust
impl std::fmt::Display for Num {
    #[inline]
    fn fmt(&self, fmtr: &mut std::fmt::Formatter) -> std::fmt::Result {
        self.0.fmt(fmtr)
    }
}
```

This could be further extended to full form:

```rust
#[derive(derive_more::Display)]
#[display(delegate, trait=Debug, field=_0.display())]
struct Path(std::path::Path, ());
```

which would end up as:

```rust
impl std::fmt::Display for Expected {
    #[inline]
    fn fmt(&self, fmtr: &mut std::fmt::Formatter) -> std::fmt::Result {
        std::fmt::Debug::fmt(&self.0.display(), fmtr)
    }
}
```

Those are further ideas though.  Even just a simple `#[display(delegate)]` which would work on structs with a single field would be neat.

tyranron:
Resolves #321

## Synopsis

See https://github.com/JelteF/derive_more/issues/321#issuecomment-1854069597:
> You’re discarding formatting flags provided by the user in format string, e.g.:
> 
> ```rust
> #[derive(derive_more::Display)]
> #[display(fmt = "{:?}", _0)]
> struct Num(usize);
> 
> impl std::fmt::Debug for Num {
>     fn fmt(&self, fmtr: &mut std::fmt::Formatter) -> std::fmt::Result {
>         self.0.fmt(fmtr)
>     }
> }
> 
> fn main() {
>     let num = Num(7);
>     println!("{num:03?}");  // prints ‘007’ as expected
>     println!("{num:03}");   // prints ‘7’ instead
> }
> ```

## Solution

See https://github.com/JelteF/derive_more/issues/321#issuecomment-1854382465:
> Theoretically, we can support this with the current syntax, because we can detect so-called _trivial_ cases and transform them into delegation (we do parse formatting string literal anyway).
> 
> ```rust
> #[derive(derive_more::Display)]
> #[display("{_0:?}")] // <--- it's clear to be a trivial delegation case
> struct Num(usize);
> ```
> 
> would expand to
> 
> ```rust
> impl std::fmt::Display for Num {
>     fn fmt(&self, fmtr: &mut std::fmt::Formatter) -> std::fmt::Result {
>         let _0 = &self.0;
>         std::fmt::Debug::fmt(_0, fmtr)
>     }
> }
> ```
> 
> rather than
> 
> ```rust
> impl std::fmt::Display for Num {
>     fn fmt(&self, fmtr: &mut std::fmt::Formatter) -> std::fmt::Result {
>         let _0 = &self.0;
>         write!(fmtr, "{_0:?}")
>     }
> }
> ```

## Checklist

[l:1]: /CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
   `#[display(fmt = "...", ("<expr>"),*)]`, and `#[display(bound(<bound>))]`
   instead of `#[display(bound = "<bound>")]`. So without the double quotes
   around the expressions and bounds.
+- The `Display` derive (and other `fmt`-like ones) now delegates to the inner type
+  when `#[display("...", (<expr>),*)]` attribute is trivially substitutable with such
+  call. ([#322](https://github.com/JelteF/derive_more/pull/322))
 - The `DebugCustom` derive is renamed to just `Debug` (gated now under a separate
   `debug` feature), and its semantics were changed to be a superset of `std` variant
   of `Debug`.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
   instead of `#[display(bound = "<bound>")]`. So without the double quotes
   around the expressions and bounds.
 - The `Debug` and `Display` derives (and other `fmt`-like ones) now transparently
-  delegates to the inner type when `#[display("...", (<expr>),*)]` attribute is
+  delegate to the inner type when `#[display("...", (<expr>),*)]` attribute is
   trivially substitutable with a transparent call.
   ([#322](https://github.com/JelteF/derive_more/pull/322))
 - The `DebugCustom` derive is renamed to just `Debug` (gated now under a separate
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,9 +25,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
   `#[display(fmt = "...", ("<expr>"),*)]`, and `#[display(bound(<bound>))]`
   instead of `#[display(bound = "<bound>")]`. So without the double quotes
   around the expressions and bounds.
-- The `Display` derive (and other `fmt`-like ones) now delegates to the inner type
-  when `#[display("...", (<expr>),*)]` attribute is trivially substitutable with such
-  call. ([#322](https://github.com/JelteF/derive_more/pull/322))
+- The `Debug` and `Display` derives (and other `fmt`-like ones) now transparently
+  delegates to the inner type when `#[display("...", (<expr>),*)]` attribute is
+  trivially substitutable with a transparent call.
+  ([#322](https://github.com/JelteF/derive_more/pull/322))
 - The `DebugCustom` derive is renamed to just `Debug` (gated now under a separate
   `debug` feature), and its semantics were changed to be a superset of `std` variant
   of `Debug`.
--- a/impl/doc/display.md
+++ b/impl/doc/display.md
@@ -11,7 +11,7 @@ the supplied format, or an automatically inferred one.
 You specify the format on each variant by writing e.g. `#[display("my val: {}", some_val * 2)]`.
 For enums, you can either specify it on each variant, or on the enum as a whole.
 
-For variants that don't have a format specified, it will simply defer to the format of the
+For variants that don't have a format specified, it will simply delegate to the format of the
 inner variable. If there is no such variable, or there is more than 1, an error is generated.
 
 
@@ -97,6 +97,52 @@ struct MyStruct<T, U, V> {
 ```
 
 
+### Delegation
+
+If the `#[display("...", args...)]` attribute is omitted, the implementation simply delegates to the format
+of the inner type, so all the additional [formatting parameters][1] do work as expected:
+```rust
+# use derive_more::Display;
+#
+#[derive(Display)]
+struct MyInt(i32);
+
+assert_eq!(format!("{:03}", MyInt(7)), "007");
+```
+
+If the `#[display("...", args...)]` attribute is specified and can be trivially substituted with a delegation
+call to the inner type, then delegation will work too:
+```rust
+# use derive_more::Display;
+#
+#[derive(Display)]
+#[display("{_0:o}")] // the same as calling `Octal::fmt()`
+struct MyOctalInt(i32);
+
+// so, additional formatting parameters do work transparently
+assert_eq!(format!("{:03}", MyInt(9)), "011");
+
+#[derive(Display)]
+#[display("{_0:02b}")] // cannot be trivially substituted with `Binary::fmt()`
+struct MyBinaryInt(i32);
+
+// so, additional formatting parameters have no effect
+assert_eq!(format!("{:07}", MyBinaryInt(2)), "10");
+```
+
+If, for some reason, delegation in trivial cases is not desired, it may be suppressed explicitly:
+```rust
+# use derive_more::Display;
+#
+#[derive(Display)]
+#[display("{}", format_args!("{_0:o}"))] // `format_args!()` opaques the inner type
+struct MyOctalInt(i32);
+
+// so, additional formatting parameters have no effect
+assert_eq!(format!("{:07}", MyInt(9)), "11");
+```
+
+
 
 
 ## Example usage
@@ -176,3 +222,8 @@ assert_eq!(UnitStruct {}.to_string(), "UnitStruct");
 assert_eq!(PositiveOrNegative { x: 1 }.to_string(), "Positive");
 assert_eq!(PositiveOrNegative { x: -1 }.to_string(), "Negative");
 ```
+
+
+
+
+[1]: https://doc.rust-lang.org/stable/std/fmt/index.html#formatting-parameters
--- a/impl/doc/display.md
+++ b/impl/doc/display.md
@@ -151,7 +151,7 @@ Or by adding [formatting parameters][1] which cause no visual effects:
 struct MyOctalInt(i32);
 
 // and so, additional formatting parameters have no effect
-assert_eq!(format!("{:07?}", MyOctalInt(9)), "11");
+assert_eq!(format!("{:07}", MyOctalInt(9)), "11");
 ```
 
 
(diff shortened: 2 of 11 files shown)


tests/display.rs
```diff
@@ -187,15 +319,15 @@ mod structs {
 
             #[derive(Display)]
             #[display(
-            "{_0} {ident} {_1} {} {}",
-            _1, _0 + _1, ident = 123, _1 = _0,
+                "{_0} {ident} {_1} {} {}",
+                _1, _0 + _1, ident = 123, _1 = _0,
```
JelteF, line 323: I think we should have tests for delegation with multi field structs too. i.e. this should trigger our delegation logic too (but on the second field)
```rust
#[derive(Display)]
#[display("{_1}")]
struct TupleDisplay(i32, i32);
```

impl/doc/display.md
```diff
@@ -11,7 +11,7 @@ the supplied format, or an automatically inferred one.
 You specify the format on each variant by writing e.g. `#[display("my val: {}", some_val * 2)]`.
 For enums, you can either specify it on each variant, or on the enum as a whole.
 
-For variants that don't have a format specified, it will simply defer to the format of the
+For variants that don't have a format specified, it will simply delegate to the format of the
 inner variable. If there is no such variable, or there is more than 1, an error is generated.
```
JelteF, line 15: I think it would be good to explicitely document this similar behaviour for Debug too. Maybe even with an example where not all fields have a custom formatting to make it extra clear.

CHANGELOG.md
```diff
@@ -25,6 +25,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
   `#[display(fmt = "...", ("<expr>"),*)]`, and `#[display(bound(<bound>))]`
   instead of `#[display(bound = "<bound>")]`. So without the double quotes
   around the expressions and bounds.
+- The `Display` derive (and other `fmt`-like ones) now delegates to the inner type
```
JelteF, line 28: I think we should explicitely call out Debug too. In the docs this is listed as a separate derive.

impl/src/fmt/mod.rs
```diff
@@ -133,28 +133,68 @@ impl ToTokens for FmtAttribute {
...
+        parsing::format(&self.lit.value())
+            .and_then(|(more, p)| {
+                // (1) There is exactly one formatting parameter.
+                more.is_empty().then_some(p)
+            })
+            .filter(|p| {
+                // (2) And the formatting parameter doesn't contain any modifiers.
+                p.spec
+                    .map(|s| {
+                        s.width.is_none() && s.precision.is_none() && s.ty.is_trivial()
```
JelteF, line 151: As far as I can tell this is missing a bunch of modifiers that are listed in the std::fmt docs, specifically:
1. Fill
2. Align
3. Sign
4. Alternate

impl/src/fmt/display.rs
```diff
@@ -222,10 +222,35 @@ impl<'a> Expansion<'a> {
     /// [`Display::fmt()`]: fmt::Display::fmt()
     fn generate_body(&self) -> syn::Result<TokenStream> {
         match &self.attrs.fmt {
-            Some(fmt) => Ok(quote! { ::core::write!(__derive_more_f, #fmt) }),
+            Some(fmt) => Ok(fmt
+                .delegatable()
+                .then(|| {
```
JelteF, line 227: nit: I think this method chain is not very easy to follow for me. I have a feeling that it might benefit readability if it is rewritten using regular match if/else.

JelteF: or maybe have the above match, match on two items: self.attrs.fmt and if the fmt is delagatable. Because I feel like there are 5 distinct cases:
1. normal formatting string
2. formatting string that is delegatable
3. no formatting string on single field variant/struct
4. no formatting string on zero-field variant/struct
5. no formatting string on multi-field variant/struct

I think it would be good to make that clear in the code and probably also add some code comments with those descriptions to each match arm.

impl/src/fmt/debug.rs
```diff
@@ -227,8 +227,30 @@ impl<'a> Expansion<'a> {
     ///
     /// [`Debug::fmt()`]: std::fmt::Debug::fmt()
     fn generate_body(&self) -> syn::Result<TokenStream> {
-        if let Some(fmt_attr) = &self.attr.fmt {
-            return Ok(quote! { ::core::write!(__derive_more_f, #fmt_attr) });
+        if let Some(fmt) = &self.attr.fmt {
+            return Ok(fmt
```
JelteF, line 231: Same as for Display derive. This stuff is really hard to follow.

impl/doc/display.md
tyranron: @JelteF already done here:
https://github.com/JelteF/derive_more/pull/322/files#diff-abe7cf81ac108e356174682bd8788a7805b865bb066505bbb0e45efec2905ee5R87-R120

CHANGELOG.md
tyranron: @JelteF since this is a CHANGELOG, and before this release there was no clear separation, I've thought that it's enough "and other `fmt`-like ones" to consider `Debug` too. But it's not a problem to mention it more explicitly.

impl/src/fmt/display.rs
tyranron: @JelteF 

> or maybe have the above match, match on two items: `self.attrs.fmt` and if the `fmt` is delegatable.

Just checking whether "`fmt` is delegatable" is not enough here, because it doesn't guarantee that the following `fmt.bindings()` extraction will have exactly one element. The way the code is composed now is that if the `fmt.bindings()` extraction for some reason returns no elements or more than one - we fall back to a normal behavior without delegation to piggyback all the compile-time checks and beautiful errors to the `write!()` macro.

The current code is the most laconic one, without bringing in fancy combinators like `itertools`. I'll think how to make it easier to follow. Since it clearly repeats for `Display` and `Debug`, the most obvious solution is to extract it to another helper method.

impl/src/fmt/mod.rs
tyranron: @JelteF good catch. I was thinking that we already parse the whole stuff in the parser, but it seems that while parser considers all the grammar correctly (we have tests for that), it does parse only the necessary minimum.

impl/src/fmt/display.rs
JelteF: tl;dr I think we should change `delagatable` to address this (if possible):
> Just checking whether "fmt is delegatable" is not enough here, because it doesn't guarantee that the following fmt.bindings() extraction will have exactly one element. 

I'd expect that these 4 checks would make sure that the bindings extraction would have exactly one element:
```rust
// (1) There is exactly one formatting parameter.
more.is_empty().then_some(p)
// ...
// (3) And either exactly one positional argument is specified.
Some(parsing::Argument::Integer(_)) | None => self.args.len() == 1,
// (4) Or the formatting parameter's name refers to some outer binding.
Some(parsing::Argument::Identifier(_)) if self.args.is_empty() => true,
// (5) Or exactly one named argument is specified for the formatting parameter's name.
Some(parsing::Argument::Identifier(name)) => {
````

Can you give an examlpe of a struct + format string where that isn't the case? And if so maybe we can add another check to `delagatable` to catch that one.

File: impl/src/fmt/mod.rs (hunk 4 above)
Comment (JelteF) [reply impl/src/fmt/mod.rs:151]: Makes sense, I guess with this PR the necessary minimum changes ;)

File: impl/src/fmt/mod.rs [hunk 7]
```diff
@@ -133,7 +133,69 @@ impl ToTokens for FmtAttribute {
...
+                .flatten()
+                .filter(|a| a.alias.as_ref().map(|a| a.0 == name).unwrap_or_default())
+                .map(|a| a.expr.clone()),
+        }?;
+
+        let trait_name = param
+            .spec
+            .map(|s| s.ty)
+            .unwrap_or(parsing::Type::Display)
+            .trait_name();
```
JelteF, line 192: When is this "default" case applicable?

tyranron: @JelteF when `format_spec` is not specified, i.e. `{}`, or `{ident}`. See [grammar for details](https://doc.rust-lang.org/stable/std/fmt/index.html#syntax).

impl/doc/display.md
```diff
@@ -176,3 +234,10 @@ assert_eq!(UnitStruct {}.to_string(), "UnitStruct");
 assert_eq!(PositiveOrNegative { x: 1 }.to_string(), "Positive");
 assert_eq!(PositiveOrNegative { x: -1 }.to_string(), "Negative");
 ```
+
+
+
+
```
JelteF, line 240: were all these newlines intentional?

impl/doc/display.md
```diff
@@ -97,6 +97,64 @@ struct MyStruct<T, U, V> {
...
+#
+#[derive(Display)]
+#[display("{_0:^o}")] // `^` is centering, but in absence of additional width has no effect
+struct MyOctalInt(i32);
+
+// and so, additional formatting parameters have no effect
+assert_eq!(format!("{:07}", MyOctalInt(9)), "11");
+```
+
+
```
JelteF, line 157: were all these newlines intentional? (feel free to keep if they were)

JelteF (approved):
LGTM. Thanks for the extensive test suite! :shipit:

impl/doc/display.md
tyranron: @JelteF yes.... 4 newlines before `##` headers and 2 newlines before `###` headers.

PR #350, opened 2024-04-18
feat(is_variant): add `#[must_use]` annotation

### LINKED ISSUE #349: `IsVariant` should annotate the generated methods with `#[must_use]`?
I'm thinking about using `IsVariant` to replace a set of manual methods I'd previously implemented:
image

I was happy to see that the methods were already `const`, but they could still be made `#[must_use]`! The reason it's in my code is because I got a pedantic clippy lint about it, but since `bool`s don't implement `Drop` and exist only to be eventually used (these static methods have no "side effects"), I think it's safe and reasonable to add this annotation! As I understand, this is what the standard library does for `Result`.

I think this would, technically, be a breaking change, so it might be good to get into 1.0, but I'm happy to quickly whip up a PR if you think this is a worthwhile addition!

Just let me know and I can get to work!

TheLostLambda:
Resolves #349 

## Synopsis & Solution

Adds a `#[must_use]` annotation to the static methods generated by `IsVariant`. I also refactored the derive to use `matches!(...)` instead of a manual `match` block with bools, but that's just a style tweak.

## Checklist

I'll be honest, I'm not 100% sure how to write tests for this one, but I have a screenshot of it in action:
Screenshot from 2024-04-18 10-32-13

[l:1]: /CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,6 +75,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 
 - The `Constructor` and `IsVariant` derives now generate `const fn` functions.
+- Static methods derived by `IsVariant` are now marked `#[must_use]`.
 - The `Unwrap` and `IsVariant` derives now generate doc comments.
 - `#[automatically_derived]` is now emitted from all macro expansions. This
   should prevent code style linters from attempting to modify the generated
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,6 +76,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 - The `Constructor` and `IsVariant` derives now generate `const fn` functions.
 - Static methods derived by `IsVariant` are now marked `#[must_use]`.
+  ([#350](https://github.com/JelteF/derive_more/pull/350))
 - The `Unwrap` and `IsVariant` derives now generate doc comments.
 - `#[automatically_derived]` is now emitted from all macro expansions. This
   should prevent code style linters from attempting to modify the generated
--- a/impl/doc/is_variant.md
+++ b/impl/doc/is_variant.md
@@ -26,13 +26,13 @@ assert!(!Maybe::<()>::Nothing.is_just());
 
 ### What is generated?
 
-The derive in the above example code generates the following code:
+The derive in the above example generates code like this:
 ```rust
 # enum Maybe<T> {
 #     Just(T),
 #     Nothing
 # }
-impl <T> Maybe<T>{
+impl<T> Maybe<T>{
     #[must_use]
     pub const fn is_just(&self) -> bool {
         matches!(self, Self::Just(..))
--- a/impl/doc/is_variant.md
+++ b/impl/doc/is_variant.md
@@ -33,11 +33,15 @@ The derive in the above example code generates the following code:
 #     Nothing
 # }
 impl <T> Maybe<T>{
+    #[inline]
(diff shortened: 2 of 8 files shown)


impl/src/is_variant.rs
```diff
@@ -45,11 +45,9 @@ pub fn expand(input: &DeriveInput, trait_name: &'static str) -> Result<TokenStre
             #[doc = #variant_name]
             #[doc = "`. Returns `false` otherwise"]
             #[inline]
+            #[must_use]
             pub const fn #fn_name(&self) -> bool {
-                match self {
-                    #enum_name ::#variant_ident #data_pattern => true,
-                    _ => false
-                }
+                matches!(self, #enum_name ::#variant_ident #data_pattern)
```
tyranron, line 50: For hygiene, this should be used like `derive_more::core::matches!`.

impl/doc/is_variant.md
```diff
@@ -33,11 +33,15 @@ The derive in the above example code generates the following code:
 #     Nothing
 # }
 impl <T> Maybe<T>{
+    #[inline]
+    #[must_use]
     pub const fn is_just(&self) -> bool {
```
tyranron, line 38: I'm not sure whether we should make an accent on `#[inline]` in docs. Let's leave only `#[must_use]` in docs.

CHANGELOG.md
```diff
@@ -75,6 +75,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 
 - The `Constructor` and `IsVariant` derives now generate `const fn` functions.
+- Static methods derived by `IsVariant` are now marked `#[must_use]`.
```
tyranron, line 78: It would be nice to also add the pull request reference, as is done below for some changes.

tyranron (changes requested):
@TheLostLambda thanks for quick reaction! 

Let's add few bikesheddings and get it merged!

tyranron:
> I'll be honest, I'm not 100% sure how to write tests for this one, but I have a screenshot of it in action:
> Screenshot from 2024-04-18 10-32-13

You can add `tests/compile_fail` with `#[deny(unused_must_use)]`. So if this will be removed in future, the test case will compile successfully, and so, fail the test.

TheLostLambda:
@tyranron Ah! Excellent! The `tests/compile_fail` is exactly what I was looking for!

I'll get to work addressing these loose ends!

TheLostLambda:
I just went a little insane trying to figure out why those tests we're running for me, but it seems the compile fail tests are all disabled on nightly Rust (only running for stable).

I think that means all of these tests are missed by CI as well, since there are a couple of failing tests unrelated to my change here.

Would you like me to add something to the CI in this PR or remove that stable-only attribute? I assume that was added in the first place to stop things from breaking when the nightly and stable compilers gave different errors, so perhaps just adding a bit in the CI is the best thing to do?

TheLostLambda:
Conclusion is that I have no idea why that last CI run didn't fail... It looks like things *are* being run in the `stable` `test` CI (one of the matrix options), but it passed in CI and failed for me, with exactly the same version of stable Rust?

I've pushed the changes that make things pass for me locally, so maybe those will break the CI version? They should?

TL;DR is that I think I've addressed all of your comments, but, unrelated, I'm not sure if the `compile_fail` tests are working correctly either in CI or on my machine...

tyranron (approved):
@TheLostLambda this is just some platform-dependent output weirdness.

Overall, now it's LGTM. Thanks!

PR #354, opened 2024-05-08
Bugfix/double reference

yfcai:
Resolves #328 

Part of #328 

## Synopsis

This is an alternative solution to the double reference problem reported in #328.

## Solution

Instead of dereferencing macro-generated bindings such as  `_0` at the use sites, do the the following:

1. In binding generation, generate `let _0 = self.0` instead of `let _0 = &self.0` in case the field `self.0` is already a reference type.

2. In body generation, add `&` before `#ident` in the call `derive_more::core::fmt::#trait_ident::fmt(#ident, __derive_more_f)`, because the function `fmt` dereferences.

3. In bounds generation, strip `&` in the bounded type for traits other than `Pointer` because the `fmt` function of those traits dereferences exhaustively.

## Open questions

1. All previous test function names are `assert`. The tests added in this PR have distinct names so that they can be executed alone via `cargo test <substring-of-test-function-name>`. Is there a reason for the naming convention `assert`? If so I'd rename the new tests.

2. Are the improvements to tests in #328 important? I can incorporate them if needed.

3. Are there any breaking changes? I'm not 100% sure.

## Checklist

- [ ] [CHANGELOG entry][l:1] is added (if required)

[l:1]: /CHANGELOG.md

### DIFF
--- a/impl/src/fmt/display.rs
+++ b/impl/src/fmt/display.rs
@@ -11,6 +11,8 @@ use crate::utils::{attr::ParseMultiple as _, Spanning};
 
 use super::{trait_name_to_attribute_name, ContainerAttributes};
 
+const POINTER_TRAIT_NAME: &str = "Pointer";
+
 /// Expands a [`fmt::Display`]-like derive macro.
 ///
 /// Available macros:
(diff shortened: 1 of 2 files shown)


impl/src/fmt/display.rs
```diff
@@ -94,8 +96,14 @@ fn expand_struct(
             .ident
             .clone()
             .map_or_else(|| syn::Member::Unnamed(i.into()), syn::Member::Named);
-        quote! {
-            let #var = &self.#member;
+        if let syn::Type::Reference(_) = f.ty {
```
JelteF, line 99: Afaict this only works for references, and will fail as soon as you use other pointer types, such as `Box<T>`.

yfcai: You are right, I will think about it more.

PR #360, opened 2024-05-19
Fix variant name interpolation in documentation for `IsVariant`

### LINKED ISSUE #357: `IsVariant` fails to interpolate variant names in documentation
My enum:

```rs
#[derive(Debug, Clone, PartialEq, IsVariant)]
pub enum Value<'a> {
    Array(Array<'a>),
    Dictionary(Dictionary<'a>),
    Boolean(bool),
    Data(Data<'a>),
    Date(Date),
    Float(f64),
    Integer(Integer),
    Real(f64),
    String(&'a str),
    Uid(Uid),
}
```

`cargo expand` shows this:

image

Which is also what appears in the generated (local) documentation:

image

Not sure if this is an edge case I've hit or just something no one has spotted previously

If you can point me in the approximate direction I can probably write & PR a fix, looks like some `quote!` tomfoolery

alpha-tango-kilo:
Resolves #357 

## Synopsis

Generated documentation is incorrect, see issue

## Solution

Ensure `stringify!` happens after `quote!` interpolation

## Checklist

---

Something for consideration: https://matklad.github.io/2022/10/24/actions-permissions.html

[l:1]: /CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -99,6 +99,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
   ([#297](https://github.com/JelteF/derive_more/pull/297))
 - Hygiene of macro expansions in presence of custom `core` crate.
   ([#327](https://github.com/JelteF/derive_more/pull/327))
+- Fix documentation of generated methods in `IsVariant` derive.
 
 ## 0.99.10 - 2020-09-11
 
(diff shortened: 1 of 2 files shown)


JelteF (approved):
Thank you for this fix. This was quite a silly bug. I don't think it's needed to add tests for this (seems hard to do for little gain).

PR #370, opened 2024-06-15
Fix clippy errors in tests

JelteF:
## Synopsis

We were only running clippy on our normal code, not tests or examples.

## Solution

This starts running clippy on the full codebase by using `--all-targets`. It
also fixes all issues that this flag found.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,7 @@ jobs:
           toolchain: stable
           components: clippy
 
-      - run: cargo clippy --workspace --all-features -- -D warnings
+      - run: cargo clippy --workspace --all-features --all-targets -- -D warnings
 
   rustfmt:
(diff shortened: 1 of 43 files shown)


tyranron (approved):
@JelteF I've taken a bit of initiative over your work and did the following adjustments:

1. Reverted removing in tests this line
   ```rust
   #[cfg(not(msrv))] // TODO: Remove once MSRV bumps 1.66 or higher.
   ```
   Since we're still on 1.65 MSRV and removal failed the CI.

2. Added brief comments to almost all `#[allow(..)]` usage in the code. This is something I've started doing recently in the crates I support, because it helps me to reconsider them much quicker when tuning lints for newer Rust versions. When I search for them in the IDE, it looks like this:  
Screenshot 2024-06-17 at 12 30 14

PR #371, opened 2024-06-18
Only put Debug-like bounds on type variables

### LINKED ISSUE #363: 'overflow evaluating the requirement' when deriving debug for item that contains itself
the following code fails to compile

```
#[derive(derive_more::Debug)]
struct Message {
    inner: Box<Message>,
}

fn main() {}
```

```
error[E0275]: overflow evaluating the requirement `Message: Debug`
 --> src/main.rs:1:10
  |
1 | #[derive(derive_more::Debug)]
  |          ^^^^^^^^^^^^^^^^^^
  |
  = note: required for `Box<Message>` to implement `Debug`
  = help: see issue #48214
  = note: this error originates in the derive macro `derive_more::Debug` (in Nightly builds, run with -Z macro-backtrace for more info)
```

the same error is reported with stuff like `Vec<Message>` and with enum instead of struct.

```
$ cargo --version --verbose
cargo 1.78.0 (54d8815d0 2024-03-26)
release: 1.78.0
commit-hash: 54d8815d04fa3816edc207bbc4dd36bf18014dbc
commit-date: 2024-03-26
host: x86_64-unknown-linux-gnu
libgit2: 1.7.2 (sys:0.18.2 vendored)
libcurl: 8.6.0-DEV (sys:0.4.72+curl-8.6.0 vendored ssl:OpenSSL/1.1.1w)
ssl: OpenSSL 1.1.1w  11 Sep 2023
os: Arch Linux Rolling Release [64-bit]
$ cat Cargo.toml
[package]
name = "derive-more-debug"
version = "0.1.0"
edition = "2021"

[dependencies]
derive_more = { version = "1.0.0-beta.6", features = ["debug"] }
```

sornas:
Resolves #363

Well, at least it's a suggestion for a resolution :p

## Synopsis

The problem, as reported in the issue, is that code like the following

```rust
#[derive(derive_more::Debug)]
struct Item {
    next: Option<Box<Item>>,
}
```

expands into something like

```rust
impl std::fmt::Debug for Item where Item: Debug { /* ... */ }
```

which does not compile. This PR changes the Debug derive so it does not emit those bounds.

## Solution

My understanding of the current code is that we iterate over all fields of the struct/enum and add either a specific
format bound (e.g. `: fmt::Binary`), a default `: fmt::Debug` bound or skip it if either it is marked
as `#[debug(skip)]` or the entire container has a format attribute. 

The suggested solution in the issue (if I understood it correctly) was to only add bounds if the type is a type
variable, since rustc already knows if a concrete type is, say, `: fmt::Debug`. So, instead of adding the bound for
every type, we first check that the type contains one of the container's type variables. Since types can be nested, it
is an unfortunately long recursive function handling the different types of types. This part of Rust syntax is probably
not going to change, so perhaps it is feasible to shorten some of the branches into `_ => false`.

One drawback of this implementation is that we iterate over the list of type variables every time we find a "leaf type".
I chose `Vec` over `HashSet` because in my experience there are only a handful of type variables per container.

## Status

I should add more tests, probably? I've at least verified the case mentioned above (for tuple structs, data structs and
enums). I must admit I haven't really looked in the documentation, but I should probably edit that too.

## Checklist

[l:1]: /CHANGELOG.md

### DIFF
--- a/impl/src/fmt/debug.rs
+++ b/impl/src/fmt/debug.rs
@@ -24,9 +24,22 @@ pub fn expand(input: &syn::DeriveInput, _: &str) -> syn::Result<TokenStream> {
         .unwrap_or_default();
     let ident = &input.ident;
 
+    let type_variables = input
+        .generics
+        .params
+        .iter()
+        .filter_map(|param| match param {
+            syn::GenericParam::Lifetime(_) => None,
+            syn::GenericParam::Type(ty) => Some(&ty.ident),
+            syn::GenericParam::Const(_) => None,
+        })
+        .collect();
+
     let (bounds, body) = match &input.data {
-        syn::Data::Struct(s) => expand_struct(attrs, ident, s, &attr_name),
-        syn::Data::Enum(e) => expand_enum(attrs, e, &attr_name),
+        syn::Data::Struct(s) => {
+            expand_struct(attrs, ident, s, type_variables, &attr_name)
+        }
+        syn::Data::Enum(e) => expand_enum(attrs, e, type_variables, &attr_name),
         syn::Data::Union(_) => {
             return Err(syn::Error::new(
                 input.span(),
@@ -64,11 +77,13 @@ fn expand_struct(
     attrs: ContainerAttributes,
     ident: &Ident,
     s: &syn::DataStruct,
+    type_variables: Vec<&syn::Ident>,
     attr_name: &syn::Ident,
 ) -> syn::Result<(Vec<syn::WherePredicate>, TokenStream)> {
     let s = Expansion {
         attr: &attrs,
         fields: &s.fields,
+        type_variables: &type_variables,
         ident,
         attr_name,
     };
@@ -99,6 +114,7 @@ fn expand_struct(
 fn expand_enum(
     mut attrs: ContainerAttributes,
     e: &syn::DataEnum,
+    type_variables: Vec<&syn::Ident>,
     attr_name: &syn::Ident,
 ) -> syn::Result<(Vec<syn::WherePredicate>, TokenStream)> {
     if let Some(enum_fmt) = attrs.fmt.as_ref() {
@@ -136,6 +152,7 @@ fn expand_enum(
             let v = Expansion {
                 attr: &attrs,
                 fields: &variant.fields,
+                type_variables: &type_variables,
                 ident,
                 attr_name,
             };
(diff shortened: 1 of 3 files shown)


impl/src/fmt/debug.rs
```diff
@@ -362,4 +393,94 @@ impl<'a> Expansion<'a> {
...
+
+                if let Some(ident) = path.get_ident() {
+                    self.type_variables
+                        .iter()
+                        .any(|type_var| *type_var == ident)
+                } else {
+                    path.segments.iter().any(|segment| {
+                        self.type_variables
+                            .iter()
+                            .any(|type_var| *type_var == &segment.ident)
```
sornas, line 414: This is outside my Rust knowledge, but I'm starting to think this is wrong. If the type is a path (like `std::vec::Vec`), then it cannot contain a type variable, correct? We only care if it is the only segment or itself a type variable (`T` in `std::vec::Vec<T>`) at some part of the path.

impl/src/fmt/debug.rs
```diff
@@ -362,4 +393,94 @@ impl<'a> Expansion<'a> {
...
+            }
+            syn::Type::Tuple(syn::TypeTuple { elems, .. }) => {
+                elems.iter().any(|ty| self.contains_type_variable(ty))
+            }
+
+            syn::Type::ImplTrait(_) => false,
+            syn::Type::Infer(_) => false,
+            syn::Type::Macro(_) => false,
+            syn::Type::Never(_) => false,
+            syn::Type::TraitObject(_) => false,
```
JelteF, line 481: Trait objects can contain type parameters:
```rust
trait MyTrait<T> {}

struct Test<T>
{
    t: Box<dyn MyTrait<T>>,
}
```

JelteF:
@sornas I have some time to work on derive_more the next ~2 weeks. Since this is one of the last items that's blocking 1.0, I was wondering if you'll have time to work on this during that time. Otherwise I think I might pick it up to get it over the finish line, if I'm able to get the other 1.0 blockers done.

sornas:
unfortunately i don't think i'll have time, so feel free to take over. but i'll let you know here if it changes.

tyranron (approved):

JelteF:
Hmm, that's not great. Can you make a new issue for this that shows some code that now fails compilation?

PR #377, opened 2024-07-02
Support _variant in outer level enum formatting for Display

### LINKED ISSUE #142: `affix` is confusing with outer-level `enum` display formatting

### LINKED ISSUE #239: ParseError when implement Display derive for enum
Awesome library, i like it!

I found some small problems during use, in some cases, it will be panic. The following is an example:

```rust
#[derive(Display)]
#[display(fmt = "{_0}")]
enum NodeType {
    Base(&'static str),
    Main(&'static str),
    N1(&'static str),
    N2(&'static str),
    N3(&'static str),
    N4(&'static str),
}
```

Here is the compile error result:
```
error: proc-macro derive panicked
  --> src/main.rs:93:10
   |
93 | #[derive(Display)]
   |          ^^^^^^^
   |
   = help: message: called `Result::unwrap()` on an `Err` value: ParseError { line: 1, column: 2, offset: 1, expected: {"}", "[0-9]", ":"} }
```

I can't find the cause of the problem from the message.

JelteF:
Resolves #142
Resolves #239

## Synopsis

This adds back support for top-level format strings of the Display derive. It
now includes the display of the variant whenever the `_variant` placeholder
apears be found. It also supports using the field

## Solution

This does not include the same support for Debug, since it is considered much less
useful there and the differences between the Debug implementation and Display
implementation make it a non-trivial port.

Only named arguments are supported in the format string, no positional ones.
This made the implementation easier, maybe in a future PR positional support
can be added.

This bumps MSRV to 1.70.0, on earlier versions the derived code throws the following error:
```rust
error: there is no argument named `_0`
    --> tests/display.rs:1373:26
     |
1373 |                 #[derive(Display)]
     |                          ^^^^^^^
     |
     = note: did you intend to capture a variable `_0` from the surrounding scope?
     = note: to avoid ambiguity, `format_args!` cannot capture variables when the format string is expanded from a macro
     = note: this error originates in the derive macro `Display` (in Nightly builds, run with -Z macro-backtrace for more info)
 ```

## Checklist

[l:1]: /CHANGELOG.md

### DIFF
--- a/clippy.toml
+++ b/clippy.toml
@@ -2,7 +2,7 @@
 # See full lints list at:
 # https://rust-lang.github.io/rust-clippy/master/index.html
 
-msrv = "1.65.0"
+msrv = "1.70.0"
 
 # Ensures consistent bracing for macro calls in the codebase.
 # Extends default settings:
--- a/clippy.toml
+++ b/clippy.toml
@@ -2,8 +2,6 @@
 # See full lints list at:
 # https://rust-lang.github.io/rust-clippy/master/index.html
 
-msrv = "1.70.0"
-
 # Ensures consistent bracing for macro calls in the codebase.
 # Extends default settings:
 # https://github.com/rust-lang/rust-clippy/blob/master/clippy_lints/src/nonstandard_macro_braces.rs#L143-L184
--- a/impl/doc/display.md
+++ b/impl/doc/display.md
@@ -26,6 +26,12 @@ The variables available in the arguments is `self` and each member of the varian
 with members of tuple structs being named with a leading underscore and their index,
 i.e. `_0`, `_1`, `_2`, etc.
 
+For enums you can also specify a shared format on the enum itself instead of
+the variant. This format is used for each of the variants, and can be
+customized per variant by including the special `{_variant}` placeholder in
+this shared format, which is then replaced by the format string that's provided
+on the variant.
+
 
 ### Other formatting traits
 
@@ -175,6 +181,7 @@ struct Point2D {
 }
 
 #[derive(Display)]
+#[display("Enum E: {_variant}")]
 enum E {
     Uint(u32),
     #[display("I am B {:b}", i)]
@@ -183,6 +190,13 @@ enum E {
     },
     #[display("I am C {}", _0.display())]
     Path(PathBuf),
+}
+
+#[derive(Display)]
+#[display("Enum E2: {_0:?}")]
+enum E2 {
+    Uint(u32),
+    String(&'static str, &'static str),
 }
 
 #[derive(Display)]
@@ -223,9 +237,11 @@ impl PositiveOrNegative {
 
 assert_eq!(MyInt(-2).to_string(), "-2");
 assert_eq!(Point2D { x: 3, y: 4 }.to_string(), "(3, 4)");
-assert_eq!(E::Uint(2).to_string(), "2");
-assert_eq!(E::Binary { i: -2 }.to_string(), "I am B 11111110");
-assert_eq!(E::Path("abc".into()).to_string(), "I am C abc");
+assert_eq!(E::Uint(2).to_string(), "Enum E: 2");
+assert_eq!(E::Binary { i: -2 }.to_string(), "Enum E: I am B 11111110");
+assert_eq!(E::Path("abc".into()).to_string(), "Enum E: I am C abc");
+assert_eq!(E2::Uint(2).to_string(), "Enum E2: 2");
+assert_eq!(E2::String("shown", "ignored").to_string(), "Enum E2: \"shown\"");
 assert_eq!(U { i: 2 }.to_string(), "Hello there!");
 assert_eq!(format!("{:o}", S), "7");
 assert_eq!(format!("{:X}", UH), "UpperHex");
--- a/impl/doc/display.md
+++ b/impl/doc/display.md
@@ -193,7 +193,7 @@ enum E {
 }
 
 #[derive(Display)]
-#[display("Enum E2: {_0:?}")]
+#[display("Enum E2: {:?}", _0)]
 enum E2 {
     Uint(u32),
     String(&'static str, &'static str),
--- a/impl/doc/display.md
+++ b/impl/doc/display.md
@@ -193,7 +193,7 @@ enum E {
 }
 
 #[derive(Display)]
-#[display("Enum E2: {:?}", _0)]
+#[display("Enum E2: {_0:?}")]
 enum E2 {
     Uint(u32),
     String(&'static str, &'static str),
--- a/impl/src/fmt/display.rs
+++ b/impl/src/fmt/display.rs
@@ -9,7 +9,7 @@ use syn::{parse_quote, spanned::Spanned as _};
 
 use crate::utils::{attr::ParseMultiple as _, Spanning};
 
-use super::{parsing, trait_name_to_attribute_name, ContainerAttributes, FmtAttribute};
+use super::{trait_name_to_attribute_name, ContainerAttributes, FmtAttribute};
 
 /// Expands a [`fmt::Display`]-like derive macro.
 ///
@@ -80,7 +80,7 @@ fn expand_struct(
     (attrs, ident, trait_ident, _): ExpansionCtx<'_>,
 ) -> syn::Result<(Vec<syn::WherePredicate>, TokenStream)> {
     let s = Expansion {
-        shared_format: None,
+        shared_attr: None,
         attrs,
         fields: &s.fields,
         trait_ident,
@@ -111,8 +111,14 @@ fn expand_struct(
 /// Expands a [`fmt`]-like derive macro for the provided enum.
 fn expand_enum(
     e: &syn::DataEnum,
-    (shared_attrs, _, trait_ident, attr_name): ExpansionCtx<'_>,
+    (container_attrs, _, trait_ident, attr_name): ExpansionCtx<'_>,
 ) -> syn::Result<(Vec<syn::WherePredicate>, TokenStream)> {
+    /*if let Some(shared_attr) = &container_attrs.fmt {
+        if shared_attr.placeholders_by_name("_variant").any(|p| false) {
+            todo!()
+        }
+    }*/
+
     let (bounds, match_arms) = e.variants.iter().try_fold(
         (Vec::new(), TokenStream::new()),
         |(mut bounds, mut arms), variant| {
@@ -135,7 +141,7 @@ fn expand_enum(
             }
 
             let v = Expansion {
-                shared_format: shared_attrs.fmt.as_ref(),
+                shared_attr: container_attrs.fmt.as_ref(),
                 attrs: &attrs,
                 fields: &variant.fields,
                 trait_ident,
@@ -196,8 +202,10 @@ fn expand_union(
 /// [`Display::fmt()`]: fmt::Display::fmt()
 #[derive(Debug)]
 struct Expansion<'a> {
-    /// Format shared between all variants of an enum.
-    shared_format: Option<&'a FmtAttribute>,
+    /// [`FmtAttribute`] shared between all variants of an enum.
+    ///
+    /// [`None`] for a struct.
+    shared_attr: Option<&'a FmtAttribute>,
 
     /// Derive macro [`ContainerAttributes`].
     attrs: &'a ContainerAttributes,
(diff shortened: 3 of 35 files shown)

### CODE REVIEW

**Comment by JelteF** [comment]
Still needs some cleanup, error-ing on edge cases, and updated docs, but the general design seems to work.

**Comment by JelteF** [comment]
> Still needs some cleanup, error-ing on edge cases, and updated docs, but the general design seems to work.

Okay all done now, this is ready for review.

File: clippy.toml [hunk 1]
```diff
@@ -2,7 +2,7 @@
 # See full lints list at:
 # https://rust-lang.github.io/rust-clippy/master/index.html
 
-msrv = "1.65.0"
+msrv = "1.70.0"
```
tyranron, line 5: I think we can remove it completely, because it defaults to `Cargo.toml`'s `rust-version` at this point.

tests/display.rs
```diff
@@ -1283,6 +1283,131 @@ mod enums {
...
             }
         }
+
+        mod shared_format {
+            use super::*;
+            mod single {
+                use super::*;
+
+                #[derive(Display)]
+                #[display("Variant: {_variant}")]
```
tyranron, line 1293: I think there should be also tests like `#[display("Variant: {}", _variant)]`

tests/compile_fail/display/shared_format_positional_placeholders.stderr
```diff
@@ -0,0 +1,17 @@
...
+2 | #[display("Stuff({})")]
+  |           ^^^^^^^^^^^
+
+error: shared format string cannot contain positional placeholders, use named placeholders instead
+ --> tests/compile_fail/display/shared_format_positional_placeholders.rs:8:11
+  |
+8 | #[display("Stuff({0})")]
+  |           ^^^^^^^^^^^^
+
+error: shared format string does not support positional placeholders, use named placeholders instead
```
tyranron, line 13: This seems to me a quite misleading error, because in tests we clearly have the valid case `#[display("Variant {_variant} {_0}")]` where the positional placeholder is used.

tyranron (changes requested):

tests/compile_fail/display/shared_format_positional_placeholders.stderr
JelteF: That's not a positional placeholder. That's a named placeholder for a tuple field. But I do understand the confusion. If you have a suggestion for better wording of the error, I'm happy to change it.

tests/display.rs
JelteF: the `_0` also doesn't work as a positional placeholder. Adding support for positional placeholders is non-trivial, because the current implementation splits the format string and calls `write!` multiple times. So if you have something like:
```rust
#[display("{}, {_variant}, {}", 123, 456)]
```
Then you need to move the expressions to the correct place.
Because this will result in the wrong output:
```rust
write!("{}, ", 123, 456)
// Write variant
write!(", {}", 123, 456)
```

I agree that positional parameters are useful, but I think we only need named parameters for 1.0. In 0.99 we didn't even support parameters at all, only the variant itself. So even with the current implementation we have a superset of the features of 0.99. And that is really all that I think we should be aiming for for 1.0 at this point. Otherwise the release might take another year.

tyranron: @JelteF so, as I see, the problem lays in the way the code desugars.

Why not try to leverage `format_args!()` and do as we usually do?
```rust
let _variant = format_args!("whatever variant formatting is", 1, 2);
write!(f, "{}, {_variant}, {}, {}", 123, 456, _variant)
```

Were there any problems with such design?

tyranron: Ahh... yes... https://github.com/rust-lang/rust/issues/92698  
Let me think...

tyranron: Okay, this desugaring just works:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=5868f867756cdddd4585eac6dff1856b

It should also simplify the implementation, as we don't need all the machinery to cut the expression to separate `write!`s.

tyranron: If you're short on time with this now, I may pick up and re-implement, if you don't mind.

JelteF: That desugaring is a lot better for sure. So yes, if you have time now, please re implement this.

tyranron (changes requested):
Needs [desugaring re-implementation](https://github.com/JelteF/derive_more/pull/377#discussion_r1688168385) as was [discussed](https://github.com/JelteF/derive_more/pull/377#discussion_r1688397549).

tyranron:
@JelteF I did the re-implementation, now it seems fine. [This restriction](https://github.com/JelteF/derive_more/pull/377/files#diff-12c6dc13d5c48b38966179cfd43412034bbe33bbbd2f00880484e4b2344f3930R117-R128) is not necessary, actually, and by doing some deeper inspection of the shared formatting attribute we may lift it in the future. This especially may be useful for the use-cases where we want to add some special modifiers (for octals, etc) for all the variants.  I haven't done it to omit complicating this PR and to not slowing the 1.0 release, as it requires more time and rich test suite.

The only thing I want to add before merging this PR is to extend tests suite a little bit more. Will do it shortly.

impl/doc/display.md
```diff
@@ -185,6 +192,13 @@ enum E {
     Path(PathBuf),
 }
 
+#[derive(Display)]
+#[display("Enum E2: {:?}", _0)]
```
JelteF, line 196: Let's use the modern syntax in the docs

```suggestion
#[display("Enum E2: {_0:?}")]
```

JelteF (commented):
Overall soooo much better than my original implementation. Left a few small notes.

impl/src/fmt/display.rs
```diff
@@ -224,70 +242,134 @@ impl<'a> Expansion<'a> {
...
+                            }
+                        }
+                    } else if self.shared_attr.is_some() {
+                        quote! { derive_more::core::format_args!(#fmt) }
+                    } else {
+                        quote! { derive_more::core::write!(__derive_more_f, #fmt) }
+                    }
+                }
+                None if self.fields.is_empty() => {
+                    let ident_str = self.ident.to_string();
```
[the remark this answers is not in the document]
tyranron: Good catch! We have [`syn::ext::IdentExt::unraw()`](https://docs.rs/syn/2.0.72/syn/ext/trait.IdentExt.html#tymethod.unraw) for this!.

impl/src/fmt/display.rs
```diff
@@ -224,70 +242,134 @@ impl<'a> Expansion<'a> {
...
+                            trait_name_to_attribute_name(self.trait_ident),
+                        ),
+                    ))
+                }
+            };
         }
+
+        if let Some(shared_fmt) = &self.shared_attr {
+            let shared_body = if let Some((shared_expr, shared_trait_ident)) =
+                shared_fmt.transparent_call()
```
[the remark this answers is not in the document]
tyranron: @JelteF ahh, yes... that's why I wanted to extend tests suite a little bit more. Good catch, again.

This issue is fixable by additionally referencing the `format_args!()` expression. Cove

> I think it would be best if we just ignored the whole shared_attr logic if the shared attr (both here in and in the big code block above). i.e. if the shared attr is `"{_variant}"` then we consider shared_attr to be `None`.

Yes, I was thinking about it too, but left "as is" as a paved way for future supporting of `{_variant}` with modifiers.

impl/src/fmt/display.rs
```diff
@@ -224,70 +242,134 @@ impl<'a> Expansion<'a> {
...
+            .map_or(true, |a| a.contains_arg("_variant"))
+        {
+            body = match &self.attrs.fmt {
+                Some(fmt) => {
+                    if let Some((expr, trait_ident)) = fmt.transparent_call() {
+                        if self.shared_attr.is_some() {
+                            let placeholder =
+                                trait_name_to_default_placeholder_literal(&trait_ident);
+
+                            quote! { derive_more::core::format_args!(#placeholder, #expr) }
```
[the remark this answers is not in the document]
tyranron: @JelteF 

> I think in my version I was actually still doing a transparent call, even when there was a shared format. It makes sense that that's not possible with your approach, and I think that is fine.

As I've [mentioned before](https://github.com/JelteF/derive_more/pull/377#issuecomment-2248649603), this is possible (to some reasonable degree, at least), by doing some deeper inspection of the shared formatting attributing and by using `&dyn fmt::Trait` for transparent calls. But this is a theme for the follow-up after-1.0-release PR.

> However, I think this code is a bit needlessly complex and by switching up the comparisons a bit it's much much easier to follow:

Makes sense.

impl/src/fmt/display.rs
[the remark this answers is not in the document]
tyranron: However, thinking about it more... one thing does not exclude another one.

tyranron (dismissed):

tyranron:
@JelteF if you don't have any more suggestions on this, I'm ready to merge this one.

tests/display.rs
```diff
@@ -1278,6 +1284,154 @@ mod enums {
...
+
+                #[derive(Display)]
+                #[display("{_variant}")]
+                enum Enum {
+                    #[display("A {_0}")]
+                    A(i32),
+                    #[display("B {}", field)]
+                    B {
+                        field: i32,
+                    },
```
JelteF, line 1325: I would want to see a transparent field as well, that correctly passes on the format options. It should 

```
// adding this variant the list
#[display("{:b}")]
Binary(i32)

// and then doing this test
assert_eq!(format("{:08}", Enum::Binary(4)), "00000100");
```

tyranron: Done.

tyranron (approved):

PR #381, opened 2024-07-03
Fix incorrect `fmt::Pointer` implementations attempt two

JelteF:
Resolves #328
Requires #377
Requires #380 

## Synopsis

`Debug` and `Display` derives allow referring fields via short syntax (`_0` for unnamed fields and `name` for named fields):
```rust
#[derive(Display)]
#[display("{_0:o}")]
struct OctalInt(i32);
```
The way this works is by introducing a local binding in the macro expansion:
```rust
let _0 = &self.0;
```

This, however, introduces double pointer indirection. For most of the `fmt` traits, this is totally OK. However, the `fmt::Pointer` is sensitive to that:
```rust
#[derive(Display)]
#[display("--> {_0:p}")]
struct Int(&'static i32);

// expands to
impl fmt::Display for Int {
    fn fmt(&self, f: fmt::Formatter<'_>) -> fmt::Result {
        let _0 = &self.0; // has `&&i32` type, not `&i32`
        write!(f, "--> {_0:p}") // so, prints address of the `_0` local binding, 
                                // not the one of the `self.0` field as we would expect
    }
}
```

## Solution

Pass all local bindings also as named parameters and dereference them there.
This allows `"{_0:p}"` to work as expected. Positional arguments and expressions
still have the previous behaviour. This seems okay IMHO, as we can explain
that in expressions these local bindings are references and that you need
to dereference when needed, such as for `Pointer`.

A downside of the current implementation is that users cannot use the names of our named parameters as names for their own named parameters, because we already use them. With some additional code this is fixable, but it doesn't seem important enough to fix. People can simply use a different name when creating their own named parameters, which is a good idea anyway because it will be less confusing to any reader of the code. If it turns out to be important to support this after all, we can still start to support it in a backwards compatible way (because now it causes a compilation failure).

## Checklist

[l:1]: /CHANGELOG.md

### DIFF
--- a/impl/doc/debug.md
+++ b/impl/doc/debug.md
@@ -25,7 +25,9 @@ once to get the address of the field itself, instead of the address of the
 reference to the field:
 
 ```rust
-#[derive(derive_more::Debug)]
+# use derive_more::Debug;
+#
+#[derive(Debug)]
 #[debug("{field:p} {:p}", *field)]
 struct RefInt<'a> {
     field: &'a i32,
@@ -43,8 +45,8 @@ are bound by respective formatting trait.
 
 E.g., for a structure `Foo` defined like this:
 ```rust
-use derive_more::Debug;
-
+# use derive_more::Debug;
+#
 #[derive(Debug)]
 struct Foo<'a, T1, T2: Trait, T3, T4> {
     #[debug("{a}")]
@@ -82,9 +84,9 @@ possible. Our aim is to avoid imposing additional bounds, as they can be added w
 In the example below, we can infer only that `V: Display`, other bounds have to be supplied by the user:
 
 ```rust
-use std::fmt::Display;
-use derive_more::Debug;
-
+# use std::fmt::Display;
+# use derive_more::Debug;
+#
 #[derive(Debug)]
 #[debug(bound(T: MyTrait, U: Display))]
 struct MyStruct<T, U, V, F> {
@@ -155,9 +157,9 @@ assert_eq!(format!("{:07?}", MyOctalInt(9)), "11");
 ## Example usage
 
 ```rust
-use std::path::PathBuf;
-use derive_more::Debug;
-
+# use std::path::PathBuf;
+# use derive_more::Debug;
+#
 #[derive(Debug)]
 struct MyInt(i32);
 
--- a/impl/doc/debug.md
+++ b/impl/doc/debug.md
@@ -14,8 +14,26 @@ This derive macro is a clever superset of `Debug` from standard library. Additio
 You supply a format by placing an attribute on a struct or enum variant, or its particular field:
 `#[debug("...", args...)]`. The format is exactly like in [`format!()`] or any other [`format_args!()`]-based macros.
 
-The variables available in the arguments is `self` and each member of the struct or enum variant, with members of tuple
-structs being named with a leading underscore and their index, i.e. `_0`, `_1`, `_2`, etc.
+The variables available in the arguments is `self` and each member of the
+struct or enum variant, with members of tuple structs being named with a
+leading underscore and their index, i.e. `_0`, `_1`, `_2`, etc. Due to
+ownership/lifetime limitations the member variables are all references to the
+fields, except when used directly in the format string. For most purposes this
+detail doesn't matter, but it is quite important when using `Pointer`
+formatting. If you don't use the `{field:p}` syntax, you have to dereference
+once to get the address of the field itself, instead of the address of the
+reference to the field:
+
+```rust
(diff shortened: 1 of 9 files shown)


tyranron (changes requested):
@JelteF could you update this PR to the latest `master`? I tried, but there are some non-trivial conflicts bound to older implementation of #377.

JelteF:
Rebased now, there's still a few more tests I want to add though.

JelteF:
Okay it turned out that the tests I wanted to add already existed

tests/debug.rs
```diff
@@ -1360,7 +1441,7 @@ mod generic {
     }
 
     #[derive(Debug)]
-    struct AliasedFieldUnnamedGenericStruct<T>(#[debug("{_0}", _0 = _1)] T, i32);
+    struct AliasedFieldUnnamedGenericStruct<T>(#[debug("{_2}", _2 = _1)] T, i32);
```
tyranron, line 1444: @JelteF I don't really like the necessity of this limitation. This is contra-intuitive for me and doesn't feel smooth from the user experience.

How about we apply the `_0 = *_0` dereferencing only for:
1. arguments requiring `Pointer`-formatting;
2. and the ones not using a defined named argument already?

So, this concrete example will expand as before. And something complex like
```rust
#[debug("{_0:p} + {_1:p} = {_0}", _0 = *_1)]
```
will expand to
```rust
write!(f, "{_0:p} + {_1:p} = {_0}", _0 = *_1, _1 = *1)
```

tyranron (changes requested):

tests/debug.rs
tyranron: If you're short on time with this, I'll do it, if you don't mind.

JelteF: I don't really see a reason to only do this for Pointer formatting only. So I'd prefer to do this change consistently.

Regarding allowing people to override the default bindings by using their name (e.g. `_0 = *_1`), I understand that argument. But honestly I think it's quite confusing to do that anyway. I mean for me parsing what your example means takes much more effort than if you'd use a name that didn't already have another meaning.

If you feel strongly about this, I'm definitely willing to accept your second suggestion. But imho, the current restriction only impacts code that's already hard to read.

tyranron: @JelteF 

> If you feel strongly about this, I'm definitely willing to accept your second suggestion.

Yup, I think I'm quite strong about it. Can imagine perfectly well how this becomes a burden in some auto-generated code.

I did the second suggestion in the https://github.com/JelteF/derive_more/pull/381/commits/842ee72734122a113291ef7e7d75983f797d082b.

impl/doc/debug.md
```diff
@@ -25,8 +45,8 @@ are bound by respective formatting trait.
 
 E.g., for a structure `Foo` defined like this:
 ```rust
-use derive_more::Debug;
-
+# use derive_more::Debug;
+#
```
JelteF, line 49: I think I prefer these `use` statements to be shown, because this is the only derive where we shadow an already existing derive. Making that clear to users is important imho.

tyranron: Ohh... `rustdoc` somehow doesn't allow to expand such examples to see the hidden code, only mdBook does. Okay, I'll rollback these.

tyranron: Done in https://github.com/JelteF/derive_more/pull/381/commits/3b67c97295bae2018700da6dd3d763c4b24a5644.

tyranron (approved):
I've extracted the similar code to the parent module and implemented the [2nd suggestion](https://github.com/JelteF/derive_more/pull/381#discussion_r1696887375) regarding expansion. 

LGTM for me now, and I'm ready to merge. Feel free to leave any notices if you still have ones.

impl/src/fmt/mod.rs
```diff
@@ -671,14 +674,17 @@ trait FieldsExt {
     /// [`FmtAttribute`] as [`FmtArgument`]s or named [`Placeholder`]s.
     ///
     /// [`syn::Ident`]: struct@syn::Ident
-    fn fmt_args_idents(&self) -> impl Iterator<Item = syn::Ident> + '_;
+    // TODO: Return `impl Iterator<Item = syn::Ident> + '_` once MSRV is bumped up to 1.75 or
+    //       higher.
```
JelteF, line 678: I'm fine with bumping MSRV to 1.75 for this

tyranron: @JelteF I'm fine too, but better do it as a separate PR then.

tyranron: I'll do shortly after merging this one.

PR #387, opened 2024-07-20
Only put Display-like bounds on type variables

### LINKED ISSUE #363: 'overflow evaluating the requirement' when deriving debug for item that contains itself

JelteF:
Only put Display-like bounds on type variables

Resolves #363
Requires #377
Related to #371

## Synopsis

The problem is that the `Display` derive adds a bounds for all types that
are used in the format string. But this is not necessary for types that
don't contain a type variable. And adding those bounds can result in
errors like for the following code:

```rust
#[derive(Display, Debug)]
#[display("{inner:?}")]
#[display(bounds(T: Display))]
struct OptionalBox<T> {
    inner: Option<Box<T>>,
}

#[derive(Display, Debug)]
#[display("{next}")]
struct ItemStruct {
    next: OptionalBox<ItemStruct>,
}
```

That code would generate the following error:

```text
error[E0275]: overflow evaluating the requirement `ItemStruct: derive_more::Display`
```

## Solution

This makes sure we don't add unnecessary bounds for Display-like
derives. It does so in the same way as #371 did for the Debug derive: By
only adding bounds when the type contains a type variable.

## Checklist

Changelog is not updated because the 0.99 did not have this issue.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -10,30 +10,16 @@ Related to <!-- paste issues/PRs references -->
 
 <!-- Remove the lines above if there are no related issues/PRs -->
 
-
-
-
 ## Synopsis
 
 <!-- Give a brief overview of the problem -->
-
-
-
 
 ## Solution
 
 <!-- Describe how exactly the problem is (or will be) resolved -->
 
-
-
-
 ## Checklist
 
 - [ ] Documentation is updated (if required)
 - [ ] Tests are added/updated (if required)
-- [ ] [CHANGELOG entry][l:1] is added (if required)
-
-
-
-
-[l:1]: /CHANGELOG.md
+- [ ] [CHANGELOG entry](/CHANGELOG.md) is added (if required)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,7 +54,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        msrv: ["1.70.0"]
+        msrv: ["1.65.0"]
         os:
           - ubuntu
           - macOS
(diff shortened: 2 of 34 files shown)


tyranron (approved):
@JelteF this seems fine. The only change I've added is extracting the duplicate code of `syn::Type`/`syn::Path` traversal to the `fmt` module shared by both expansions.

PR #394, opened 2024-08-06
Remove nightly cfg from impl crate

JelteF:
We can use `rustversion` for this. For some of the tests its still
needed because of this error:

```
error[E0658]: non-inline modules in proc macro input are unstable
  --> tests/error/mod.rs:53:1
   |
53 | mod nightly;
   | ^^^^^^^^^^^^
   |
   = note: see issue #54727 <https://github.com/rust-lang/rust/issues/54727> for more information
```


JelteF:
Hmm, okay apparently I cannot use `rustversion` in doc tests (which is where we need it). So I'll leave this open to look again later.

PR #406, opened 2024-08-29
Swap macro re-export modules (#405)

### LINKED ISSUE #405: Stop exporting traits by default and early 2.0 release
The amount of opened issues and negative feedback raised because we export traits together with derives by default in 1.0 is much larger than I had expected. So I'm left wondering if we made the right decision here, and if the downsides are worth the upsides. 

The downsides I'm seeing are:

1. Confusing people that add derive_more to an existing codebase with strange import conflict errors
2. Conflicting with people calling their types Error
3. Causing people upgrading from 0.99 extra breaking changes

Benefits:
1. No need to import the trait if you use it

I think the following reddit comment made good points on why this is the possibly the wrong tradeoff. tl;dr people familiar with rust are used to the errors of not having imported an std trait, and rust-analyzer helps solve these issues easily. But these conflicting import errors have are uncommon, and thus a hassle to solve especially because there's no tooling to fix them automatically.

>  I have to agree, reexporting std here is the baaaad kind of efficiency. You're creating unexpected, non-standard behavior. And behavior that existing tooling (errors, lints, etc.) does not handle gracefully (because it's abnormal). That's almost always bad, and creates invisible debt and the worst kind of work (fake work trying to deal with obtuse, non-substantive errors). And in this case the only plus is that it removes the need to do something rust analyzer and an ide would almost automate anyway. I love the derive_more crate, but I agree that this is the bad kind of complexity. (e.g. I *expect* to see ` use std:...` if I'm using it) Any introduction of unexpected behavior needs to solve a serious problem. An example of headache: I was just writing some exploratory code using the new derive_more and also just spent time trying to figure out where naming collisions were coming from. It doesn't matter that the "finished" code wouldn't have those naming collisions. When I'm in the process of writing code I'm going to explore things based on assumptions about the state of imports etc in a file/crate. (Perhaps there's a large upside that I'm not seeing, but this mostly seems to be doing the work that tooling already does in a way that's opaque. Also, this probably seems trivial to you because you're intimately familiar with your own crate. Almost no one else is going to expect this, creating headache whenever it comes up. Edge case or otherwise.) 
>  ——
>
> Edit: I love your crate and am super thankful for it!!! Just to be clear.   (Realized that sentiment may be lost in the minutae discussion.)

Source: https://www.reddit.com/r/rust/comments/1emi4fi/comment/lhkmjpk/

Right now I'm definitely leaning towards the feeling of: "If I could go back in time, I think we should not have made this change".

The question then becomes, how can we make changing this back as painless as possible, now that we have made this "mistake".

The best approach I can see is by doing the following:
1. Create a 2.0 release. To not cause silent breakage for people already on 1.0.
2. Create a module like `derive_more::both` or `derive_more::trait_derive` that people already on 1.0 can use.
3. Do this relatively quickly, while 1.0 is not used a lot yet.

Related to:
- https://github.com/JelteF/derive_more/issues/401
- https://github.com/JelteF/derive_more/issues/271#issuecomment-2290003199
- https://www.reddit.com/r/rust/comments/1emi4fi/comment/lh0jdm2/

tyranron:
Resolves #405

As was decided in #405, the crate's root should re-export only macros. Re-exporting of macros along with their corresponding traits should be moved to `with_trait` module.

## Checklist

- [x] Tests are added/updated

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -66,17 +66,6 @@ pub mod __private {
     pub use crate::vendor::thiserror::aserror::AsDynError;
 }
 
-/// Module containing macro definitions only, without corresponding traits.
-///
-/// Use it in your import paths, if you don't want to import traits, but only macros.
-pub mod derive {
-    // This can be unused if no feature is enabled. We already error in that case, but this warning
-    // distracts from that error. So we suppress the warning.
-    #[allow(unused_imports)]
-    #[doc(inline)]
-    pub use derive_more_impl::*;
-}
-
 // The modules containing error types and other helpers.
 
 #[cfg(feature = "add")]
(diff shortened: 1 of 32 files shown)


JelteF:
Needs a rebase after merging: https://github.com/JelteF/derive_more/pull/410

src/lib.rs
```diff
@@ -119,303 +108,316 @@ mod try_unwrap;
...
-        mod $new_module_name {
+// This can be unused if no feature is enabled. We already error in that case, but this warning
+// distracts from that error. So we suppress the warning.
+#[allow(unused_imports)]
+#[doc(inline)]
+pub use derive_more_impl::*;
+
+/// Module containing macro definitions with their corresponding traits along.
+///
+/// Use it in your import paths, if you do want to import macros along with their traits.
```
JelteF, line 119:
```suggestion
/// Module containing derive definitions with their corresponding traits along.
///
/// Use it in your import paths, if you do want to import derives along with their traits.
```

src/lib.rs
```diff
@@ -66,17 +66,6 @@ pub mod __private {
...
-/// Module containing macro definitions only, without corresponding traits.
-///
-/// Use it in your import paths, if you don't want to import traits, but only macros.
-pub mod derive {
-    // This can be unused if no feature is enabled. We already error in that case, but this warning
-    // distracts from that error. So we suppress the warning.
-    #[allow(unused_imports)]
-    #[doc(inline)]
-    pub use derive_more_impl::*;
-}
```
JelteF, line 78: Should we be removing this module? I feel like it's a somewhat unnecessary backwards compatibility break. At the very least this removal should be documented in the changelog.

JelteF (dismissed):
Feel free to merge after you fix the above issues

tyranron:
@JelteF

> Feel free to merge after you fix the above issues

I'd like to, but it's impossible without your approval at the moment, as I cannot approve PRs created by myself.

JelteF (approved):

PR #415, opened 2024-09-27
Add new derive: AsVariant

### LINKED ISSUE #358: `AsVariant` for enum
It would be nice to have a derive macro that is similar to `IsVariant` and `Unwrap` but that turns an enum into an `Option<T>` (where `T` is the data of that variant).

crazymerlyn:
Resolves #358 

## Synopsis

A new derive for generating `as_foo`, `as_bar` methods for an enum with fields `foo` and `bar`.

## Solution

Similar to TryUnwrap but generates functions returning an Option instead of a Result.

## Checklist

### DIFF
--- a/tests/as_variant.rs
+++ b/tests/as_variant.rs
@@ -1,0 +1,106 @@
+#![cfg_attr(not(feature = "std"), no_std)]
+#![cfg_attr(nightly, feature(never_type))]
+#![allow(dead_code)] // some code is tested for type checking only
+
+use derive_more::AsVariant;
+
+#[derive(AsVariant)]
+enum Either<TLeft, TRight> {
+    Left(TLeft),
+    Right(TRight),
+}
+
+#[derive(AsVariant)]
+#[derive(Debug, PartialEq)]
+#[as_variant(ref, ref_mut)]
+enum Maybe<T> {
+    Nothing,
+    Just(T),
+}
+
+#[derive(AsVariant)]
+enum Color {
+    Rgb(u8, u8, u8),
+    Cmyk(u8, u8, u8, u8),
+}
+
+/// With lifetime
+#[derive(AsVariant)]
+enum Nonsense<'a, T> {
+    Ref(&'a T),
+    NoRef,
+    #[as_variant(ignore)]
+    NoRefIgnored,
+}
+
+#[derive(AsVariant)]
+enum WithConstraints<T>
+where
+    T: Copy,
+{
+    One(T),
+    Two,
+}
+
+#[derive(AsVariant)]
+enum KitchenSink<'a, 'b, T1: Copy, T2: Clone>
+where
+    T2: Into<T1> + 'b,
+{
+    Left(&'a T1),
+    Right(&'b T2),
+    OwnBoth(T1, T2),
+    Empty,
+    NeverMind(),
+    NothingToSeeHere(),
+}
+
+/// Single variant enum
+#[derive(AsVariant)]
+enum Single {
+    Value(i32),
+}
+
+#[derive(AsVariant)]
+#[derive(Debug, PartialEq)]
+#[as_variant(ref, ref_mut)]
+enum Tuple<T> {
+    None,
+    Single(T),
+    Double(T, T),
+    Triple(T, T, T),
+}
+
+#[test]
+pub fn test_as_variant() {
+    assert_eq!(Maybe::<()>::Nothing.as_nothing(), Some(()));
+    assert_eq!(Maybe::Just(1).as_just_ref(), Some(&1));
+    assert_eq!(Maybe::Just(42).as_just_mut(), Some(&mut 42));
+
+    assert_eq!(Maybe::<()>::Nothing.as_just(), None);
+    assert_eq!(Maybe::Just(1).as_nothing_ref(), None);
+    assert_eq!(Maybe::Just(42).as_nothing_mut(), None);
+}
+
+#[test]
+pub fn test_as_variant_mut() {
+    let mut value = Tuple::Double(1, 12);
+
+    if let Some((a, b)) = value.as_double_mut() {
+        *a = 9;
+        *b = 10;
+    }
+
+    assert_eq!(value, Tuple::Double(9, 10));
+}
+
+#[cfg(nightly)]
+mod never {
+    use super::*;
+
+    #[derive(AsVariant)]
+    enum Enum {
+        Tuple(!),
+        TupleMulti(i32, !),
+    }
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
   its trait along is possible now via `use derive_more::with_trait::SomeTrait`.
   ([#406](https://github.com/JelteF/derive_more/pull/406))
 
+### Added
+
+- Derive for `AsVariant`.
+
 ### Fixed
 
 - Associated types of type parameters not being treated as generics in `Debug`
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,6 +53,7 @@ default = ["std"]
 add = ["derive_more-impl/add"]
 add_assign = ["derive_more-impl/add_assign"]
 as_ref = ["derive_more-impl/as_ref"]
+as_variant = ["derive_more-impl/as_variant"]
 constructor = ["derive_more-impl/constructor"]
 debug = ["derive_more-impl/debug"]
 deref = ["derive_more-impl/deref"]
@@ -124,6 +125,11 @@ required-features = ["as_ref"]
 name = "as_ref"
 path = "tests/as_ref.rs"
 required-features = ["as_ref"]
+
+[[test]]
+name = "as_variant"
+path = "tests/as_variant.rs"
+required-features = ["as_variant"]
 
 [[test]]
 name = "boats_display_derive"
--- a/impl/Cargo.toml
+++ b/impl/Cargo.toml
@@ -52,6 +52,7 @@ default = []
 add = []
 add_assign = []
 as_ref = ["syn/extra-traits", "syn/visit"]
+as_variant = ["dep:convert_case"]
 constructor = []
 debug = ["syn/extra-traits", "dep:unicode-xid"]
 deref = []
@@ -78,6 +79,7 @@ full = [
     "add",
     "add_assign",
     "as_ref",
+    "as_variant",
     "constructor",
     "debug",
     "deref",
--- a/impl/doc/as_variant.md
+++ b/impl/doc/as_variant.md
@@ -1,0 +1,70 @@
+# What `#[derive(AsVariant)]` generates
+
+When an enum is decorated with `#[derive(AsVariant)]`, for each variant `foo` in
+the enum, with fields `(a, b, c, ...)`, a public instance method `as_foo(self) -> Option<(a, b, c, ...)>` is generated.
+If you don't want the `as_foo` method generated for a variant you can put the
+`#[as_variant(ignore)]` attribute on that variant.
+If you want to treat a reference, you can put the `#[as_variant(ref)]` attribute on the enum declaration or that variant, then `as_foo_ref(self) -> Option<(&a, &b, &c, ...)>` will be generated. You can also use mutable references by putting `#[as_variant(ref_mut)]`.
+
+
+
+
+## Example usage
+
+```rust
+# use derive_more::AsVariant;
+#
+#[derive(AsVariant)]
+#[as_variant(ref)]
+enum Maybe<T> {
+    Just(T),
+    Nothing
+}
+
+assert_eq!(Maybe::<()>::Nothing.as_nothing(), Some(()));
+assert_eq!(Maybe::<()>::Nothing.as_just(), None);
+assert_eq!(Maybe::Just(1).as_just(), Some(1));
+assert_eq!((&Maybe::Just(42)).as_just_ref(), Some(&42));
+```
+
+
+### What is generated?
+
+The derive in the above example generates code like this:
+```rust
+# enum Maybe<T> {
+#     Just(T),
+#     Nothing
+# }
+impl<T> Maybe<T>{
+    #[must_use]
+    pub fn as_just(self) -> Option<(T)> {
+        match self {
+            Maybe::Just(field_0) => Some((field_0)),
+            _ => None,
+        }
+    }
+    #[must_use]
+    pub fn as_just_ref(&self) -> Option<(&T)> {
+        match self {
+            Maybe::Just(field_0) => Some((field_0)),
+            _ => None,
+        }
+    }
+
+    #[must_use]
+    pub fn as_nothing(self) -> Option<()> {
+        match self {
+            Maybe::Nothing => Some(()),
+            _ => None,
+        }
+    }
+    #[must_use]
+    pub fn as_nothing_ref(&self) -> Option<()> {
+        match self {
+            Maybe::Nothing => Some(()),
+            _ => None,
+        }
+    }
+}
+```
--- a/impl/src/as_variant.rs
+++ b/impl/src/as_variant.rs
@@ -1,0 +1,135 @@
+use crate::utils::{AttrParams, DeriveType, State};
+use convert_case::{Case, Casing};
+use proc_macro2::TokenStream;
+use quote::{format_ident, quote};
+use syn::{DeriveInput, Fields, Result, Type};
+
+pub fn expand(input: &DeriveInput, trait_name: &'static str) -> Result<TokenStream> {
+    let state = State::with_attr_params(
+        input,
+        trait_name,
+        "as_variant".into(),
+        AttrParams {
+            enum_: vec!["ignore", "owned", "ref", "ref_mut"],
+            variant: vec!["ignore", "owned", "ref", "ref_mut"],
+            struct_: vec!["ignore"],
+            field: vec!["ignore"],
+        },
+    )?;
+    assert!(
+        state.derive_type == DeriveType::Enum,
+        "AsVariant can only be derived for enums",
+    );
+
+    let enum_name = &input.ident;
+    let (imp_generics, type_generics, where_clause) = input.generics.split_for_impl();
+
+    let variant_data = state.enabled_variant_data();
+
+    let mut funcs = vec![];
+    for (variant_state, info) in
+        Iterator::zip(variant_data.variant_states.iter(), variant_data.infos)
+    {
+        let variant = variant_state.variant.unwrap();
+        let fn_name = format_ident!(
+            "as_{}",
+            variant.ident.to_string().to_case(Case::Snake),
+            span = variant.ident.span(),
+        );
+        let ref_fn_name = format_ident!(
+            "as_{}_ref",
+            variant.ident.to_string().to_case(Case::Snake),
+            span = variant.ident.span(),
+        );
+        let mut_fn_name = format_ident!(
+            "as_{}_mut",
+            variant.ident.to_string().to_case(Case::Snake),
+            span = variant.ident.span(),
+        );
+        let variant_ident = &variant.ident;
+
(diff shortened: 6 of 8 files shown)


tests/as_variant.rs
```diff
@@ -0,0 +1,106 @@
...
+
+#[test]
+pub fn test_as_variant() {
+    assert_eq!(Maybe::<()>::Nothing.as_nothing(), Some(()));
+    assert_eq!(Maybe::Just(1).as_just_ref(), Some(&1));
+    assert_eq!(Maybe::Just(42).as_just_mut(), Some(&mut 42));
+
+    assert_eq!(Maybe::<()>::Nothing.as_just(), None);
+    assert_eq!(Maybe::Just(1).as_nothing_ref(), None);
+    assert_eq!(Maybe::Just(42).as_nothing_mut(), None);
```
joshka, line 82: Misses a few cases (and rearranged a little):

```suggestion
    assert_eq!(Maybe::<()>::Nothing.as_nothing(), Some(()));
    assert_eq!(Maybe::<()>::Nothing.as_nothing_ref(), Some(()));
    assert_eq!(Maybe::<()>::Nothing.as_nothing_mut(), Some(()));

    assert_eq!(Maybe::<()>::Nothing.as_just(), None);
    assert_eq!(Maybe::<()>::Nothing.as_just_ref(), None);
    assert_eq!(Maybe::<()>::Nothing.as_just_mut(), None);

    assert_eq!(Maybe::Just(1).as_nothing(), None);
    assert_eq!(Maybe::Just(1).as_nothing_ref(), None);
    assert_eq!(Maybe::Just(1).as_nothing_mut(), None);

    assert_eq!(Maybe::Just(42).as_just(), Some(42));
    assert_eq!(Maybe::Just(42).as_just_ref(), Some(&42));
    assert_eq!(Maybe::Just(42).as_just_mut(), Some(&mut 42));
```

joshka (approved):
This is a feature that I want to see. It looks similar to the unwrap and try_unwrap derives and seems like it implements the principle of least surprise.

In terms of releasing this, would it be possible to release this on a 1.x version instead of waiting for a 2.0 release (if that's going to be delayed)?

I noticed a few minor things that could be improved.

tyranron (changes requested):
I feel quite uncomfortable about this feature, because of the following reasons:

1. `#[derive(AsVariant)]` doesn't add any new functionality to the library. It just fully mirrors the `#[derive(TryUnwrap)]` with the only exception of using `Option` instead of `Result` in method signatures, which is insignificant, since we always can just add `.ok()` if we don't want a `Result`. What is the purpose of it, then? A whole new macro just for another naming? Is this worth it?

2. Regarding the naming itself, there are semantic problems. Frankly, `#[derive(TryUnwrap)]` doesn't have them. However, in Rust, by best practices and some common ecosystem conventions, `as_*`, `into_*` and `to_*` method prefixes are expected to point to the concrete method signatures:
    ```rust
    // `ref` to `ref` conversion
    fn as_something(&self) -> PossibleContainer<&Something>;
    // `mut ref` to `mut ref` conversion
    fn as_mut_something(&mut self) -> PossibleContainer<&mut Something>;
    // `ref` to value conversion
    fn to_something(&self) -> PossibleContainer<Something>;
    // infallible value to value conversion
    fn into_something(&self) -> Something;
    // fallible value to value conversion
    fn try_into_something(&self) -> Container<Something>;
    ```
    And so, seeing `fn as_something(self) -> Option<Something>` method just doesn't feel right for me.  
    I would intuitive expect the  `#[derive(AsVariant)]` macro to only generate the following methods:
    ```rust
    fn as_something(&self) -> Option<&Something>;
    fn as_mut_something(&mut self) -> Option<&mut Something>;
    ```
    Just because `As`/`as_` is expected to be about reference-to-reference semantics only.

Given the above, I'm quite unsure in which direction to go:
- We may reduce the `#[derive(AsVariant)]`  macro scope to reference-to-reference semantics only, but then, again, it's covered by `#[try_unwrap(ref, ref_mut)]` already. Should we duplicate the feature, or remove the `#[try_unwrap(ref, ref_mut)]` from the `#[derive(TryUnwrap)]`?
- We may split  `#[derive(TryUnwrap)]` into `#[derive(AsVariant)]` + `#[derive(IntoVariant)]` which are same but return `Option`. Is it worth it?
- We may not accept  `#[derive(AsVariant)]` at all, but folks raise concerns that `#[derive(TryUnwrap)]` is unintuitive.

@JelteF I would appreciate any thoughts on this from you.

------------------------------

> This is a feature that I want to see. It looks similar to the unwrap and try_unwrap derives and seems like it implements the principle of least surprise.

@joshka could you elaborate more on this? Why `TryUnwrap` doesn't sastisfy your needs?

joshka:
> However, in Rust, by best practices and some common ecosystem conventions, as_*, into_* and to_* method prefixes are expected to point to the concrete method signatures:

There are several examples of returning Option from an as_method in the std lib. Click a few types and look for as_ methods.

https://doc.rust-lang.org/std/primitive.char.html#method.as_ascii
```
pub const fn as_ascii(&self) -> Option<AsciiChar>
```

I do however think you might be right though despite that example. https://rust-lang.github.io/api-guidelines/naming.html#c-conv suggests as_ is for Borrowed -> Borrowed. The various cell types have `into_inner` methods that return `Option<T>`. I don't think I've see try_ methods that return Options too often though (I think that was implied by your last example, but I'm not 100% certain).

@joshka could you elaborate more on this? Why TryUnwrap doesn't sastisfy your needs?

The specific place that I want these is in the events enums in the crossterm events module (link above).

Conversion from an enum which isn't the anticipated type has no value to me having any extra information that the Error adds. Option conveys the information succinctly and will better represent the reality of what is happening than an error. The error fundamentally is that this is not the thing you asked for. There's nothing more to add ever.

Put another way, there's a very real semantic difference between enums where you're expecting a certain variant at some time (try_unwrap) vs one where you are taking some action only if it is some variant (as_variant). In the places where I want to use this, the latter is closer to what I want the library to convey and not the former.

> Given the above, I'm quite unsure in which direction to go:...

I don't think it's worth removing TryUnwrap, but it's good to give people reasonable Options (see what I did there ;)). I'm not wedded to a particular as_ naming for that however. If you're curious, perhaps take a look at the various places the existing derives are used https://github.com/search?q=derive_more+tryunwrap&type=code perhaps there's some additional insight in how each of those cases end up calling the derived methods?

joshka:
>it's a bit silly that we would only allow them to generate a function that returns a Result, requiring them to add .ok() everywhere.

The search link above confirms that this is a thing that happens:

E.g.
https://github.com/search?q=repo%3Arestatedev%2Frestate%20try_unwrap&type=code

(these are many lines from the same file:)
```rust
            assert_that!(record.is_data(), eq(true));
            let data = record.try_unwrap_data().unwrap();
            let original: String = data.decode().unwrap();
                assert_that!(record.is_filtered_gap(), eq(true));
                let gap = record.try_unwrap_filtered_gap().unwrap();
                assert_that!(gap.to, eq(LogletOffset::new(5)));
                let data = record.try_unwrap_data().unwrap();
                let gap = record.try_unwrap_filtered_gap().unwrap();
                let data = record.try_unwrap_data().unwrap();
                let gap = record.try_unwrap_trim_gap().unwrap();
                let data = record.try_unwrap_data().unwrap();
        let gap = record.try_unwrap_trim_gap().unwrap();
```
This sort of pattern of `try_unwrap_*().unwrap()` is found in a few of the results.


Or https://github.com/search?q=repo%3Atangramdotdev%2Ftangram%20try_unwrap&type=code
Which has many instances of the same pattern:
```rust
		};
			let object = if let Some(subpath) = &referent.subpath {
				let directory = object
					.try_unwrap_directory()
					.ok()
					.ok_or_else(|| tg::error!("expected a directory"))?;
				directory.get(&handle, subpath).await?.into()
```

joshka:
Rust 1.83 that [came out today](https://blog.rust-lang.org/2024/11/28/Rust-1.83.0.html#stabilized-apis) has a couple of stabilized functions `ControlFlow::break_value() -> Option<B>` and ControlFlow::continue_value() ->  that are basically this. They're suffixed `_value`. `Result` has just `ok(()` and `err()`, which might also be instructive on naming. To me this suggests that perhaps entirely dropping the idea of a prefix altogether and instead allow the user to add a suffix that is meaningful to them would be the right choice. This avoids possible conflicts with methods which are used to create variants (e.g. `std::io::Error::other(...)` is a constructor, but would conflict without some prefix/suffix).

tyranron:
> Only the naming really remains then. I think `as_variant` is an acceptable naming scheme. I agree it's not super obvious, but I also don't think it's terrible. And it's nice and short, and fits well with `is_variant`.

@JelteF fine, then. If we're okay with naming and the design, let's polish and merge it.

------------------------------------

> To me this suggests that perhaps entirely dropping the idea of a prefix altogether and instead allow the user to add a suffix that is meaningful to them would be the right choice. This avoids possible conflicts with methods which are used to create variants (e.g. `std::io::Error::other(...)` is a constructor, but would conflict without some prefix/suffix).

@joshka @JelteF given the above, I would like to propose the following design, which seems the least confusing to me:

```rust
#[derive(AsVariant)] // `ref` semantics by default (the most common case)
enum Maybe<T> {
    Nothing, // units are ignored by default, since has no sense
    Just(T),
    Pair { a: u8, b: T },
}
// Expands to:
fn as_just(&self) -> Option<&T> {}
fn as_pair(&self) -> Option<(&u8, &T)> {}

// We can tune each variant separately:
#[derive(AsVariant)]
enum Maybe<T> {
    #[as_variant(owned, ref_mut)]
    Nothing,
    #[as_variant(owned, ref)]
    Just(T),
    #[as_variant(ignore)]
    Pair { a: u8, b: T },
}
// Expands to:
fn to_nothing(self) -> Option<()> {}
fn as_mut_nothing(&mut self) -> Option<&mut ()> {}
fn to_just(self) -> Option<T> {}
fn as_just(&self) -> Option<&T> {}

// Or all of them together:
#[derive(AsVariant)]
#[as_variant(owned, ref, ref_mut)]
enum Maybe<T> {
    Nothing, // still ignored by default
    Just(T),
    #[as_variant(owned)]
    Pair { a: u8, b: T }, // variant-level stuff takes precedence
}
// Expands to:
fn to_just(self) -> Option<T> {}
fn as_just(&self) -> Option<&T> {}
fn as_mut_just(&mut self) -> Option<&mut T> {}
fn to_pair(self) -> Option<(u8, T)> {}
```

This way, in my opinion, we have reasonable defaults in semantics (without generating likely unnecessary code) and conventional naming by default. While still have the ability to opt-in other behavior if it's required.

Regarding the naming, we may additionally allow renaming methods (in a separate PR, though):
```rust
#[derive(AsVariant)]
#[as_variant(owned = "into_{name}", ref, ref_mut = "as_ref_mut_{name}")]
enum Maybe<T> {
    Nothing, // still ignored by default
    Just(T),
    #[as_variant(owned, ref_mut = "as_{name}_mut")]
    Pair { a: u8, b: T }, // variant-level stuff takes precedence
}
// Expands to:
fn into_just(self) -> Option<T> {}
fn as_just(&self) -> Option<&T> {}
fn as_ref_mut_just(&mut self) -> Option<&mut T> {}
fn to_pair(self) -> Option<(u8, T)> {}
fn as_pair_mut(&mut self) -> Option<(&mut u8, &mut T)> {}
```

And, another thought just slipped into my mind:  we only need an additional knob to cover `TryUnwrap` here too.
```rust
#[derive(AsVariant)]
#[as_variant(owned = "into_{name}", ref, ref_mut = "as_ref_mut_{name}")]
#[as_variant(result)] // default is `option`
enum Maybe<T> {
    Nothing, // still ignored by default
    #[as_variant(option)]
    Just(T), // variant-level stuff takes precedence
    #[as_variant(owned, ref_mut = "as_{name}_mut")]
    Pair { a: u8, b: T }, // variant-level stuff takes precedence
}
// Expands to:
fn into_just(self) -> Option<T> {}
fn as_just(&self) -> Option<&T> {}
fn as_ref_mut_just(&mut self) -> Option<&mut T> {}
fn to_pair(self) -> Result<(u8, T), Error> {}
fn as_pair_mut(&mut self) -> Result<(&mut u8, &mut T), Error> {}
```
Assuming that users would like to have `Result` much rarer than `Option`, we could deprecate `TryUnwrap` altogether, since `AsVariant` covers it with little additional effort.

JelteF:
@tyranron thanks for the detailed example. That's super helpful to get a sense of how this works out.

I did some more research, and I agree that in the ecosystem `as_variant` generally seems to return an `Option<&T>`. I think for owned (`Option<T>`) I think the `into_variant` prefix does not seem super common though. It seems more common to simply use the snake_case version of the variant name, e.g. `ok()` and `err()`, but also `left()` and `right()`.

What the default ownership for the derive should be I'm not sure. But given how all the interested parties so far seem to want the owned behaviour, I'd say that that is a good default. If we do that though, and we go with the names I propose above, then maybe we shouldn't call the derive `AsVariant`, but simply `Variant`.

tyranron:
@JelteF so just to see the whole thing, I'll describe your alterations to the proposed design. Please, correct me if anything is wrong:

```rust
#[derive(Variant)] // `owned` semantics by default (the most common case)
enum Maybe<T> {
    Nothing, // units are ignored by default, since has no sense
    Just(T),
    Pair { a: u8, b: T },
}
// Expands to:
fn just(self) -> Option<T> {}
fn pair(self) -> Option<(u8, T)> {}

// We can tune each variant separately:
#[derive(Variant)]
enum Maybe<T> {
    #[variant(owned, ref_mut)]
    Nothing,
    #[variant(owned, ref)]
    Just(T),
    #[variant(ignore)]
    Pair { a: u8, b: T },
}
// Expands to:
fn nothing(self) -> Option<()> {}
fn as_mut_nothing(&mut self) -> Option<&mut ()> {}
fn just(self) -> Option<T> {}
fn as_just(&self) -> Option<&T> {}

// Or all of them together:
#[derive(Variant)]
#[variant(owned, ref, ref_mut)]
enum Maybe<T> {
    Nothing, // still ignored by default
    Just(T),
    #[variant(owned)]
    Pair { a: u8, b: T }, // variant-level stuff takes precedence
}
// Expands to:
fn just(self) -> Option<T> {}
fn as_just(&self) -> Option<&T> {}
fn as_mut_just(&mut self) -> Option<&mut T> {}
fn pair(self) -> Option<(u8, T)> {}

// With custom renaming:
#[derive(Variant)]
#[variant(owned = "into_{name}", ref, ref_mut = "as_ref_mut_{name}")]
enum Maybe<T> {
    Nothing, // still ignored by default
    Just(T),
    #[variant(owned, ref_mut = "as_{name}_mut")]
    Pair { a: u8, b: T }, // variant-level stuff takes precedence
}
// Expands to:
fn into_just(self) -> Option<T> {}
fn as_just(&self) -> Option<&T> {}
fn as_ref_mut_just(&mut self) -> Option<&mut T> {}
fn into_pair(self) -> Option<(u8, T)> {}
fn as_pair_mut(&mut self) -> Option<(&mut u8, &mut T)> {}

// With returning `Result`:
#[derive(Variant)]
#[variant(result)] // default is `option`
enum Maybe<T> {
    Nothing, // still ignored by default
    #[variant(option)]
    Just(T), // variant-level stuff takes precedence
    #[variant(owned, ref_mut)]
    Pair { a: u8, b: T },
}
// Expands to:
fn just(self) -> Option<T> {}
fn pair(self) -> Result<(u8, T), Error> {}
fn as_mut_pair(&mut self) -> Result<(&mut u8, &mut T), Error> {}
```

And my thoughts about it:

> I think the `into_variant` prefix does not seem super common though. It seems more common to simply use the snake_case version of the variant name, e.g. `ok()` and `err()`, but also `left()` and `right()`.

@joshka pointed out above, just the snake_case version of the variant name is quite often used for variant constructors (e.g. `fn just(value: impl Into<u8>) -> Maybe<u8>`) and I would like to avoid such collisions by default. I agree that `into_variant` prefix is not super common, and because `into` naming quite collides with `Into` trait, I often expect there infallible conversion, which is not the case. That's why I proposed the default naming as `to_variant` for owned conversions.

> But given how all the interested parties so far seem to want the owned behaviour, I'd say that that is a good default.

I agree that's a very often used feature, however, I'm unsure that `owned` case is more common than `ref` case. At least in my codebases, I see the tendency to have `ref` case more often.

> If we do that though, and we go with the names I propose above, then maybe we shouldn't call the derive `AsVariant`, but simply `Variant`.

Very seductive, however, I do think that people will intuitively look for `AsVariant` (the created issue is a somewhat proof), and `Variant` may be not that obvious from its name for unfamiliar users.

joshka:
TL;DR: suffixes / prefixes should be additional configuration, not there by default.

A good way to frame why this is needed is that it implements getters for an enum. Because the enum can be valid as any of the variants, the getters must return `Option<T>`.

QQ: are there existing derive_more attributes where the prefix / suffix are configurable already?

The ordering of suffixes in the comment above is generally a bit wrong. Based on idiomatic ordering, `_ref` / `_mut` should generally be at the end not in the middle of the function name. Background:

- https://rust-lang.github.io/rfcs/0199-ownership-variants.html
- https://github.com/rust-lang/rfcs/blob/master/text/0344-conventions-galore.md#suffix-ordering
- https://github.com/ratatui/ratatui/pull/1084#discussion_r1606044341 (where we discussed suffix ordering in some Ratatui code related to opts)

> As @joshka pointed out above, just the snake_case version of the variant name is quite often used for variant constructors (e.g. fn just(value: impl Into<u8>) -> Maybe<u8>) and I would like to avoid such collisions by default. I agree that into_variant prefix is not super common, and because into naming quite collides with Into trait, I often expect there infallible conversion, which is not the case. That's why I proposed the default naming as to_variant for owned conversions.

I think not prefixing by default might make more sense than doing this to avoid constructors. It fits better with https://rust-lang.github.io/api-guidelines/naming.html#c-getter

> I agree that's a very often used feature, however, I'm unsure that owned case is more common than ref case. At least in my codebases, I see the tendency to have ref case more often.

I don't have a good gauge of which would be most common, but rather than looking at it from that perspective, I look at the above and see a lot of complexity in the rules about naming. A much simpler alternative which seems intuitive to me is to make the default `owned` (same default as `TryInto`) be unprefixed, and add additional `ref` and `mut` options which affect the suffix giving:

```rust
// #[variant(owned)
fn just(self) -> Option<T>
fn pair(self) -> Option<(u8, T)>

// #[variant(ref)
fn just_ref(&self) -> Option<&T>
fn pair_ref(&self) -> Option<&(u8, T)>

// #[variant(mut)
fn just_mut(&mut self) -> Option<&mut T>
fn pair_mut(&mut self) -> Option<&mut (u8, T)>
```

This seems much more predictable and intuitve, with their still being the option to enhance this with prefixes if needed, while the opposite of removing the prefixes seems like a chore for when you there is no conflict with constructors.

> Assuming that users would like to have Result much rarer than Option, we could deprecate TryUnwrap altogether, since AsVariant covers it with little additional effort.

TryUnwrap still has its place, and removing it would mean that the owned/ref/mut config gets twice as large in the variant case.

I think I'd still call this `AsVariant`, `ToVariant`,`GetVariant`, or even `Getters` even though it's not producing `as_` / `to_` methods (I'd lean towards `ToVariant` or `GetVariant`). This naming seems to communicate better what the derive macro will do when casually reading code more-so than `Variant`. The methods produced really are just getters that avoid panics / errors. This is a much weaker preference than the ones above though (perhaps this is 60% sure rather than 90%?).

Diegovsky:
As a minor comment: I think the default should be `ref` if it is named `AsVariant`. I don't like the name `Variant` for not being easily understandable on its own. `AsVariant` communicates very nicely it is a convertion-like derive, while also being similat to `IsVariant`.

However, fell free to ignore this comment. I feel my previous comment is actually actionable and kills most bike shedding.

TheNeikos:
Is this still being worked on? And if so, is the design decided in the end? I wouldn't mind picking it up if that's the case.

PR #425, opened 2024-11-10
Make strings in error types public

okamt:
## Synopsis

Can't access strings in error types, useful for mapping to another error type or if you want to change the Display impl

## Solution

Make strings in the dedicated error type structs (FromStrError, TryIntoError etc.) public.

## Checklist

none are applicable, I think


tyranron:
@okamt those are private not out of nothing, but for the ability to toss them around without introducing breaking changes to the library API. Could you elaborate more on your use case? Give some example?

okamt:
Oh, I just realized the error doesn't store the wrong string at all... lol

PR #440, opened 2025-01-19
Fix unsized fields usage in `derive(Display)` (#432)

### LINKED ISSUE #432: Unsized tail fields in derive(Display)
```rust
use derive_more::Display;

#[derive(Display, Debug)]
#[display("{head}.{tail}")]
struct Struct {
    head: char,
    tail: str,
}

fn main() {
    let dat = [51i32, 3028017];
    let s = unsafe { &*(&raw const dat as *const [i32] as *const Struct) };
    println!("{:?}", s);
    println!("{}", s);
}
```

The derived Debug impl prints `Struct { head: '3', tail: "14" }` but the Display impl does not compile.

```console
error[E0277]: the size for values of type `str` cannot be known at compilation time
   --> src/main.rs:3:10
    |
3   | #[derive(Display, Debug)]
    |          ^^^^^^^ doesn't have a size known at compile-time
4   | #[display("{head}.{tail}")]
    |                   ------ required by a bound introduced by this call
    |
    = help: the trait `Sized` is not implemented for `str`
note: required by an implicit `Sized` bound in `derive_more::core::fmt::rt::Argument::<'_>::new_display`
   --> $RUSTUP_HOME/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/fmt/rt.rs:113:24
    |
113 |     pub fn new_display<T: Display>(x: &T) -> Argument<'_> {
    |                        ^ required by the implicit `Sized` requirement on this type parameter in `Argument::<'_>::new_display`
    = note: this error originates in the macro `$crate::format_args` which comes from the expansion of the derive macro `Display` (in Nightly builds, run with -Z macro-backtrace for more info)
```

tyranron:
Resolves #432

## Synopsis

See https://github.com/JelteF/derive_more/issues/432#issue-2761083580:
> ```rust
> use derive_more::Display;
> 
> #[derive(Display, Debug)]
> #[display("{head}.{tail}")]
> struct Struct {
>     head: char,
>     tail: str,
> }
> 
> fn main() {
>     let dat = [51i32, 3028017];
>     let s = unsafe { &*(&raw const dat as *const [i32] as *const Struct) };
>     println!("{:?}", s);
>     println!("{}", s);
> }
> ```
> 
> The derived Debug impl prints `Struct { head: '3', tail: "14" }` but the Display impl does not compile.
> 
> ```
> error[E0277]: the size for values of type `str` cannot be known at compilation time
>    --> src/main.rs:3:10
>     |
> 3   | #[derive(Display, Debug)]
>     |          ^^^^^^^ doesn't have a size known at compile-time
> 4   | #[display("{head}.{tail}")]
>     |                   ------ required by a bound introduced by this call
>     |
>     = help: the trait `Sized` is not implemented for `str`
> note: required by an implicit `Sized` bound in `derive_more::core::fmt::rt::Argument::<'_>::new_display`
>    --> $RUSTUP_HOME/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/fmt/rt.rs:113:24
>     |
> 113 |     pub fn new_display<T: Display>(x: &T) -> Argument<'_> {
>     |                        ^ required by the implicit `Sized` requirement on this type parameter in `Argument::<'_>::new_display`
>     = note: this error originates in the macro `$crate::format_args` which comes from the expansion of the derive macro `Display` (in Nightly builds, run with -Z macro-backtrace for more info)
> ```

The problem is introduced by #381, and its cause is that the [`FmtAttribute::additional_deref_args()`](https://github.com/JelteF/derive_more/blob/v1.0.0/impl/src/fmt/mod.rs#L281-L300) implementation is too aggressive and is enabled for any `fmt` trait, not just `fmt::Pointer`.
```rust
#[allow(unreachable_code)]
#[automatically_derived]
impl derive_more::core::fmt::Display for Struct2 {
    fn fmt(&self, __derive_more_f: &mut derive_more::core::fmt::Formatter<'_>) -> derive_more::core::fmt::Result {
        let head = &self.head;
        let tail = &self.tail;
        derive_more::core::write!(__derive_more_f, "{head}.{tail}", head = *head, tail = *tail)
    }
}
```

## Solution

Apply `FmtAttribute::additional_deref_args()` only for `fmt::Pointer` placeholders.

## Checklist

- [x] Tests are added/updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Top-level `#[display("...")]` attribute on an enum not working transparently
   for directly specified fields.
   ([#438](https://github.com/JelteF/derive_more/pull/438))
+- Incorrect dereferencing of unsized fields in `Debug` and `Display` expansions.
+  ([#440](https://github.com/JelteF/derive_more/pull/440))
 
 
 ## 1.0.0 - 2024-08-07
--- a/impl/src/fmt/mod.rs
+++ b/impl/src/fmt/mod.rs
@@ -306,7 +306,9 @@ impl FmtAttribute {
 
     /// Returns an [`Iterator`] over the additional formatting arguments doing the dereferencing
     /// replacement in this [`FmtAttribute`] for those [`Placeholder`] representing the provided
-    /// [`syn::Fields`] and requiring it
+    /// [`syn::Fields`] and requiring it ([`fmt::Pointer`] ones).
+    ///
+    /// [`fmt::Pointer`]: std::fmt::Pointer
     fn additional_deref_args<'fmt: 'ret, 'fields: 'ret, 'ret>(
         &'fmt self,
         fields: &'fields syn::Fields,
@@ -314,7 +316,7 @@ impl FmtAttribute {
         let used_args = Placeholder::parse_fmt_string(&self.lit.value())
             .into_iter()
             .filter_map(|placeholder| match placeholder.arg {
-                Parameter::Named(name) => Some(name),
+                Parameter::Named(name) if placeholder.trait_name == "Pointer" => Some(name),
                 _ => None,
             })
             .collect::<Vec<_>>();
--- a/impl/src/fmt/mod.rs
+++ b/impl/src/fmt/mod.rs
@@ -316,7 +316,9 @@ impl FmtAttribute {
         let used_args = Placeholder::parse_fmt_string(&self.lit.value())
             .into_iter()
             .filter_map(|placeholder| match placeholder.arg {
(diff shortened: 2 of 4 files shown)


tyranron:
@JelteF after merging this, we can finally cut the 2.0.0 release! 🎉

tyranron:
ping @JelteF

tyranron:
ping @JelteF

JelteF (approved):

tyranron:
@JelteF let's cut the 2.0.0 release?

JelteF:
yes sounds good. Can you prepare a PR?

PR #443, opened 2025-02-02
impl rename_all for #[derive(Display)] (#216)

### LINKED ISSUE #216: Serde like `rename` for Display & FromStr trait on enum

ModProg:
Resolves #216 
## Checklist


impl/src/fmt/display.rs
```diff
@@ -148,14 +153,18 @@ fn expand_enum(
...
+            let attrs = attrs.map(Spanning::into_inner).unwrap_or_default();
+
             let ident = &variant.ident;
 
             if attrs.fmt.is_none()
                 && variant.fields.is_empty()
                 && attr_name != "display"
+                && container_attrs.rename_all.is_none()
             {
                 return Err(syn::Error::new(
```
tyranron, line 169: @ModProg hmmm... I don't see why `rename_all` argument should allow implicit naming for other traits rather than `Display` only? Could you elaborate on that?

For me, the situation `assert_eq!(format!("{:o}", Enum::VariantOne), "variant_one");` seems quite awkward.

impl/src/fmt/mod.rs
```diff
@@ -509,16 +570,47 @@ impl Placeholder {
...
+#[derive(Debug)]
+struct ContainerAttributes<T> {
     /// Interpolation [`FmtAttribute`].
     fmt: Option<FmtAttribute>,
 
     /// Addition trait bounds.
     bounds: BoundsAttribute,
+
+    /// Rename unit enum variants following a similar behavior as [`serde`](https://serde.rs/container-attrs.html#rename_all).
+    rename_all: Option<T>,
```
tyranron, line 582: @ModProg instead of modifying the `ContainerAttributes` with an additional slot (which is useless for the `Debug`), I think it's better to leave it "as is", and:
1. Move `RenameAllAttribute` into `display` module.
2. Inside `display` module do:
    ```rust
    struct ContainerAttributes {
        rename_all: RenameAllAttribute,
        common: super::ContainerAttributes,
    }
    ```
    and parse it as `Either<RenameAllAttribute, super::ContainerAttributes>` to omit repeating parsing of the `super::ContainerAttributes` twice.

tyranron (changes requested):

impl/src/fmt/display.rs
ModProg: My idea was you could use different display traits for different ways of `rename_all` if you want to.

ModProg: I have seen crates use different format traits for stuff like this in the past, but I don't mind removing the functionality. i.e. make `rename_all` only available for `Display`.

impl/src/fmt/mod.rs
ModProg: I thought It'd make sense to allow to have `rename_all` and `bounds` in the same attribute. But that will only at most save one line anyway.

impl/src/fmt/display.rs
JelteF: I agree that this is weird: `assert_eq!(format!("{:o}", Enum::VariantOne), "variant_one");`

But I also think `assert_eq!(format!("{:o}", Enum::VariantOne), "VariantOne");` is weird. And as I understand that's what we do currently. I would be happier to see the `Octal` derive simply fail for unit enum variants without an `#[octal(...)]` attribute. Instead of defaulting to stringifying them in the same way as we do by default for `Display`, but that would be a breaking change. And doing another major release to fix that doesn't seem worth it.

So my vote would be, let's do whatever is easiest code wise. And let's create an issue that we add to the 3.0 milestone, to remove this weird behaviour completely then.

tyranron: @JelteF 

> And as I understand that's what we do currently.

Nope, at the moment, we [forbid implicit naming for other traits rather than `Display`](https://github.com/JelteF/derive_more/blob/v2.0.1/impl/src/fmt/display.rs#L156-L165). That's why the question was risen.

tyranron (approved):
@ModProg good work! Thanks!

klardotsh:
Any chance a stable(ish?) release with this diff is coming in the upcoming future? This is super useful stuff for making clean JSON API wrappers that take lowercased string constants as arguments.

tyranron:
@klardotsh I'm planning to do the release upon finishing the milestone: https://github.com/JelteF/derive_more/milestone/4

I'm actively working on it right now.

klardotsh:
Awesome news, thanks! And thanks for your work on making all this stuff even more ergonomic to use!

tisonkun:
Do we have a `rename` attr (not `rename_all`) to specify the rename target explicitly?

I often find`rename_all` attr a footgun, see:

For `HTTPRequest`

* serde -> h_t_t_p_request
* display -> httprequest
* sqlx -> http_request

Not sure what flavor derive_more adopts :P

PR #445, opened 2025-02-03
Add Rust Playground metadata

JelteF:
derive_more is being pulled into the playground due to being a
dependency of some other crate in the top 100 crates, probably
<https://docs.rs/selectors>.

This adds metadata so that it gets installed with all features once
selectors updates to derive_more 2.0


tyranron (approved):

JelteF:
Opened a PR for selectors here: https://github.com/servo/stylo/pull/117

PR #453, opened 2025-03-26
Fix compiling error on wasm32-none

jasl:
## Synopsis

When compiling on `wasm32-none` target, it fails with the error
```
  error[E0463]: can't find crate for `std`
    |
    = note: the `wasm32v1-none` target may not support the standard library
    = note: `std` is required by `derive_more` because it does not declare `#![no_std]`

  error: cannot find macro `include_str` in this scope
    --> /Users/jasl/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/derive_more-1.0.0/src/lib.rs:42:11
     |
  42 |     doc = include_str!("../README.md")
     |           ^^^^^^^^^^^
     |
  help: consider importing one of these macros
     |
  53 + use crate::core::include_str;
     |
  53 + use core::include_str;
```

## Solution

Follow the help suggestion

## Checklist

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -39,7 +39,7 @@
         feature = "from",
         feature = "into"
     ),
-    doc = include_str!("../README.md")
+    doc = core::include_str!("../README.md")
 )]
 #![cfg_attr(not(feature = "std"), no_std)]
 #![cfg_attr(docsrs, feature(doc_auto_cfg))]


CryZe:
That is not the fix. You forgot to disable std.

jasl:
> That is not the fix. You forgot to disable std.

I've tested it, and it works. I had this trouble when I was trying to upgrade `polkadot-sdk`.

I'm not a Rust expert, but the significant difference between `wasm32-unknown-unknown` and `wasm32-none` (introduced since 1.78) is that `wasm32-none` doesn't have `std` crate at all, which causes the error.

References:
- <https://doc.rust-lang.org/beta/rustc/platform-support/wasm32v1-none.html>
- <https://doc.rust-lang.org/beta/rustc/platform-support/wasm32v1-none.html>

CryZe:
This should probably be reported to the Rust compiler instead. `include_str` should work just fine even without std.

jasl:
> This should probably be reported to the Rust compiler instead. `include_str` should work just fine even without std. It seems like you broke something internal pretty hard. There is no bug in this crate.

I don't think so.
The error message clearly says to use `include_str!` from the `core` crate; you have to call it explicitly.
There is no bug on the Rust compiler side. To fix it, we only need to follow the suggestion.

I have tested it, and it works. If you don't believe it, you can reproduce the issue yourself.

tyranron (approved):
@jasl good catch! Thanks for the PR!

CryZe:
I explicitly tested that this is not necessary, but whatever, I guess we just merge it anyway 🤷

tyranron:
@jasl @CryZe it would be nice to report this to the compiler if one could craft an MRE.

PR #454, opened 2025-03-31
 fix: Ignore deprecation warnings on generated code

### LINKED ISSUE #419: Hide deprecation warnings inside derive expansions
Consider the following example:
```rust
use derive_more::Display;
#[derive(Debug, Display)]
pub enum Test {
    #[deprecated(note = "this is deprecated")]
    B { x: usize },
}
```

Compiling this produces a warning:
```
warning: use of deprecated field `derive_more_deprecate::Test::B::x`: this is deprecated
 --> src/derive_more_deprecate.rs:5:9
  |
5 |     B { x: usize },
  |         ^
```

Adding `#[allow(deprecated)]` does not help since the attribute does not get added to the `impl Display`.

This also happens with other derive macros (`From`, `Error`, ...).

aborgna-q:
Resolves #419

## Synopsis

Deprecating a field or variant doesn't produced un-ignorable warnings in the code generated  by derive_more (the only solution was to globally `#![allow(deprecated)]` in the whole module).

## Solution

Adds `allow(deprecated)` to the generated code.

## Checklist

- [x] ~Documentation is updated (if required)~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
   implicit naming of unit enum variants or unit structs when deriving `Display`.
   ([#443](https://github.com/JelteF/derive_more/pull/443))
 
+### Fixed
+
+- Ignore deprecation warnings on generated code. ([#454](https://github.com/JelteF/derive_more/pull/454))
+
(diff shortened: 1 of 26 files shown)


tyranron (approved):
@aborgna-q thank you! ❤️

PR #459, opened 2025-04-24
Support optional `Error::source` (#426)

### LINKED ISSUE #426: Optional source for Error
I’d like to be able to use `derive_more` with optional sources, like:

```rs
#[derive(derive_more::Error, derive_more::Display, Debug)]
#[display("it’s an error")]
struct MyErr {
    source: Option<Box<dyn std::error::Error + Send + Sync + 'static>>,
}
```

Notably, this does work with `thiserror`.

At the moment this gives:

```
error[E0599]: the method `as_dyn_error` exists for enum `Option<Box<dyn Error + Send + Sync>>`, but its trait bounds were not satisfied
    |
187 | #[derive(derive_more::Error, derive_more::Display, Debug)]
    |          ^^^^^^^^^^^^^^^^^^
    |
571 | pub enum Option<T> {
    | ------------------ doesn't satisfy `_: AsDynError<'_>` or `_: Error`
    |
    = note: the following trait bounds were not satisfied:
            `std::option::Option<std::boxed::Box<dyn derive_more::Error + std::marker::Send + std::marker::Sync>>: derive_more::Error`
            which is required by `std::option::Option<std::boxed::Box<dyn derive_more::Error + std::marker::Send + std::marker::Sync>>: derive_more::__private::AsDynError<'_>`
note: the method `as_dyn_error` exists on the type `std::boxed::Box<(dyn derive_more::Error + std::marker::Send + std::marker::Sync + 'static)>`
    |
9   |     fn as_dyn_error(&self) -> &(dyn Error + 'a);
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    = note: this error originates in the derive macro `derive_more::Error` (in Nightly builds, run with -Z macro-backtrace for more info)
```

tyranron:
Resolves #426

## Synopsis

See https://github.com/JelteF/derive_more/issues/426#issue-2688714756:
> I’d like to be able to use `derive_more` with optional sources, like:
> ```rust
> #[derive(derive_more::Error, derive_more::Display, Debug)]
> #[display("it’s an error")]
> struct MyErr {
>     source: Option<Box<dyn std::error::Error + Send + Sync + 'static>>,
> }
> ```

## Solution

The way, the `thiserror` does it, is by [checking](https://docs.rs/thiserror-impl/2.0.12/src/thiserror_impl/expand.rs.html#50) the [type name to contain `Option`](https://docs.rs/thiserror-impl/2.0.12/src/thiserror_impl/expand.rs.html#559-583). Relying on type could be fragile, as the call site could introduce its own type named `Option` in the scope where macro is called. Ideally, we need to aid this case with type machinery.

For supporting this via type machinery, we obviously need some sort of specialization to specialize case `Option<E: Error>` over `E: Error`, since vanilla Rust coherence naturally hits the wall of "upstream crates may add a new impl of trait `std::error::Error` for type `Option<_>` in future versions". The usual solution is to use [autoref-based specialization](https://lukaskalbertodt.github.io/2019/12/05/generalized-autoref-based-specialization.html) as [we do for `AsRef`](https://docs.rs/derive_more/2.0.1/src/derive_more/as.rs.html). However, autoref-based specialization doesn't work in a generic context (i.e. `MyError<E1, E2>`), since it requires trait bounds to work, and we cannot describe conditional trait bounds in Rust, which leads to the situation that this case could not be supported at all:
```rust
struct MyError<E> {
    source: Option<E>,
}
```

So, choosing between checking the field's type syntactically (and so loosing support for renamed types) and not supporting optional `source` in generic contexts, I would definitely choose to support generics despite having naming issues in edge cases. As the result, this PR implements the similar syntactically-based solution as the `thiserror` crate does. The implications of this are warned and described in the documentation.

## Future possibilities

Additionally, we may support `#[error(source(optional))]` attribute, which will allow using renamed types naturally:
```rust
use derive_more::{Display, Error};

struct Simple;

#[display("Oops!")]
struct Generic(#[error(source(optional))] RenamedOption<Simple>); // forces recognizing as `Option`
type RenamedOption<T> = Option<T>;
```
However, it's better to be implemented as a separate PR, to not overcomplicate this one.

## Checklist

- [x] Tests are added/updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Support `#[display(rename_all = "<casing>")]` attribute to change output for
   implicit naming of unit enum variants or unit structs when deriving `Display`.
   ([#443](https://github.com/JelteF/derive_more/pull/443))
+- Support `Option` fields for `Error::source()` in `Error` derive.
+  ([#459](https://github.com/JelteF/derive_more/pull/459))
 
 ### Fixed
 
--- a/impl/doc/error.md
+++ b/impl/doc/error.md
@@ -52,6 +52,38 @@ Backtraces don't work though, because the `Backtrace` type is only available in
 `std`.
 
 
+### `Option`al fields
+
+Deriving `source()` is supported naturally for `Option<_>`-typed fields.
+
+> **WARNING**: Field type is **considered as `Option<_>` only syntactically**,
+> meaning that:
+> - Neither renaming `std::option::Option` would work:
+>   ```rust,compile_fail
+>   # use derive_more::{Display, Error};
+>   #
+>   #[derive(Debug, Display, Error)]
+>   struct Simple;
+>
+>   #[derive(Debug, Display, Error)]
+>   #[display("Oops!")]
+>   struct Generic(RenamedOption<Simple>); // not recognized as `Option`
+>   type RenamedOption<T> = Option<T>;
+>   ```
+> - Nor using any custom type named as `Option` would:
+>   ```rust,compile_fail
+>   # use derive_more::{Display, Error};
+>   #
+>   # #[derive(Debug, Display, Error)]
+>   # struct Simple;
+>   #
+>   #[derive(Debug, Display, Error)]
+>   #[display("No way!")]
+>   struct Generic(Option<Simple>);        // cannot use `?` syntax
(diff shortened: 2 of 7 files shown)


tyranron:
ping @JelteF

JelteF (approved):

PR #467, opened 2025-05-06
Support `#[from_str(rename_all = "...")]` attribute (#216)

### LINKED ISSUE #216: Serde like `rename` for Display & FromStr trait on enum

tyranron:
Resolves #216   
Requires #468, #469

## Synopsis

We have `#[display(rename_all = "...")]` attribute for flat enums. The counterpart for `FromStr` derive is missing.

## Solution

This PR adds support `#[from_str(rename_all = "...")]` attribute on enums. Once this attribute is specified, the matching is always exact (for now, by default, matching could be fuzzy if there is no similarly named variants).

## Additionally

This PR fully reworks `FromStr` derive implementation to avoid using old `utils` machinery.

Adds support for unit structs (like `struct Unit;`) for the counterpart behavior of `Display`.

Adds `tests/compile_fail/from_str`.

## Checklist

- [x] Tests are added/updated


tyranron:
@JelteF this PR has grown a little bit too much than I've expected, so I've dissected some of it logical parts into separate PRs to keep things simpler:
- Reworking `FromStr` derive in new style (#468)
- Supporting flat expansion for structs without fields (as `Display` does)
- Supporting `#[from_str(rename_all = "...")]` on top level and variant level (this PR)

tyranron:
ping @JelteF

JelteF (approved):

PR #468, opened 2025-05-07
Rework `FromStr` derive

tyranron:
Required for #467

This PR fully reworks `FromStr` derive implementation to avoid using old `utils` machinery.

Also, adds `compile_fail` tests for `FromStr`.

## Checklist

- [x] ~~Documentation is updated~~ (not required)
- [x] Tests are added/updated
- [x] ~~[CHANGELOG entry](/CHANGELOG.md) is added~~ (not required)

PR #469, opened 2025-05-07
Support flat structs in `FromStr`

tyranron:
Required for #467  
Requires #468

## Synopsis

We support flat structs for `Display`:
```rust
#[derive(Display)]
struct Foo;
```
But `FromStr` doesn't support it.

## Solution

Allow using `#[derive(FromStr)]` on structs with no fields.

Similarly to enums it will use fuzzy matching. Exact matching would be possible to enable in #469 by using `#[from_str(rename_all = "...")]` attribute.

## Checklist

- [x] Tests are added/updated


tyranron:
ping @JelteF

JelteF (approved):
Pretty useless feature on its own, but if it's a building block for something more useful that seems fine.

PR #472, opened 2025-05-22
Implemented `#[skip]` for the `Add` derive

### LINKED ISSUE #437: Ignore zero-width fields like `PhantomData`
Ignore zero-width fields to allow complier hinting with types like `PhantomData` while deriving trait imlementations.
```rs
// This should work.
#[derive(derive_more::Add)]
struct A<B>(u16, PhantomData<B>);

// Equivalent to:
impl<B> Add for A<B> {
    type Output = Self;

    fn add(self, rhs: Self) -> Self::Output {
        Self(self.0 + rhs.0, PhantomData)
    }
}
```
This behavior can be seen in [`#[repr(transparent)]`](https://doc.rust-lang.org/nomicon/other-reprs.html#reprtransparent), where the annotation will work as long as there is only one non-zero-width type.

ErmitaVulpe:
Resolves #437

## Synopsis

As outlined in #437 it isn't possible to derive `Add` for structs containing zero sized fields

Example:
```rust
#[derive(Add)] // method not found in `PhantomData<T>`
struct TupleWithZst<T>(i32, PhantomData<T>);

#[derive(Add)] // method not found in `PhantomData<T>`
struct StructWithZst<T> {
    x: i32
    _marker: PhantomData<T>,
}
```

## Solution

Added a `#[skip]` attribute that allows to skip zero sized fields like so:
```rust
#[derive(Add)]
struct TupleWithZst<T>(i32, #[skip] PhantomData<T>);

#[derive(Add)]
struct StructWithZst<T> {
    x: i32,
    #[skip]
    _marker: PhantomData<T>,
}
```

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
   ([#459](https://github.com/JelteF/derive_more/pull/459))
 - Support structs with no fields in `FromStr` derive.
   ([#469](https://github.com/JelteF/derive_more/pull/469))
+- Support `#[skip]` attribute for the `Add` derive to allow skipping zero sized
+  fields. ([#472](https://github.com/JelteF/derive_more/pull/472))
 
 ### Changed
 
--- a/impl/doc/add.md
+++ b/impl/doc/add.md
@@ -76,6 +76,26 @@ The behaviour is similar for more or less fields.
 
 
 
+## Skipping fields
+Sometimes the struct needs to hold a zero sized field, most commonly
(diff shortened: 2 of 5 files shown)


ErmitaVulpe:
To do this I needed to get a bit more familiar with the codebase and i noticed that `Mul` and `MulAssign` depended on `add_helpers.rs` for handling of the `#[mul(forward)]` attrib so I also implemented it for them :P

Also I changed the way that trait bounds are added to the generics so that the generics contained in the skipped fields do not require the derived trait to be implemented for them.

tyranron (approved):

PR #473, opened 2025-05-26
Bootstrap `derive(PartialEq)` (#163)

tyranron:
Part of #163

## Synopsis

This PR is the initial part of #163 and adds only structural expansion of `derive(PartialEq)`.

## Solution

Mirrors the `PartialEq` expansion of `std`, but applies trait bounds for generics correctly.

## Checklist

- [x] Documentation is updated
- [x] Tests are added/updated
- [x] [CHANGELOG entry](/CHANGELOG.md) is added

Issue #163, opened 2021-06-09
Feature idea: `PartialEq` and `PartialOrd` for the wrapped type


CGMossa:
I'm finding myself needing to wrap a type "quickly", but then I don't want to have any
specific restrictions on the inner-type, other than being wrapped.

`PartialOrd` and `PartialEq` where `Rhs = InnerType` is a small ergonomic advantage if at all possible?

I see that #51 was shortly dismissed, but maybe this is a good time to maybe revisit this.

I have also been thinking if this is possible to add through a trait, but I think this is less explicit then
what can be done using a derive-macro. Also the object-safety issue comes up if I want to do that..

JelteF:
Maybe it's indeed time to revisit, what do you think @tyranron?

JelteF:
I'm fine with merging a PR that provides this.

CGMossa:
This is very good to hear. Unfortunately, I'm a little busy right now, but please, if no-one has taken a stab at this after 2 months, then ping me. I'm very interested in contributing to `derive_more` development.

tyranron:
[Roadmap](https://github.com/JelteF/derive_more/pull/473#issuecomment-2910403957) for implementing this:
- [ ] Bootstrap `derive(PartialEq)` with structural expansion. (#473)
- [ ] Implement `PartialEq::ne()` method (`std` omits it) to allow better performance for cases when it's optimized.
- [ ] Support `#[partial_eq(skip)]` attribute.
- [ ] Support `#[partial_eq(where(<bounds>))]` attribute (required when fields may contain private types).
- [ ] Implement forwarding expansion for newtypes (single-field structs) with `#[partial_eq(forward)]` support.
- [ ] Support `#[partial_eq(<types>)]` to allow the case requested in this issue.
- [ ] Implement `derive(Eq)` expanding as `impl <T> Eq for Type<T> where Self: PartialEq {}`.

JelteF:
Regarding `derive(Eq)`: I'm wondering whether we should make that simply also derive `PartialEq`, at least by default. One of my personal pet peeves with the default derive is that you always need type out both.

tyranron:
[...]

PR #475, opened 2025-05-28
Generate `PartialEq::ne()` (#163)

tyranron:
Part of #163  
Follows #473

## Synopsis

`derive(PartialEq)` in `std` generates only `PartialEq::eq()` method implementation in its expansion. This silently kills any potential performance benefits if the underlying types implement  `PartialEq::ne()` more efficiently than the default implementation (`!PartialEq::eq()`).

## Solution

Generate `PartialEq::ne()` implementation always as well, by structurally calling `!=` operator instead of `==`.

## Additionally

Unfortunately, the `assert_ne!()` macro also doesn't call `PartialEq::ne()` or `!=` operator inside its expansion, and only negates the equality check. Found this when running tests against a purposely incorrect  `PartialEq::ne()` implementation. That's why the `assert_ne!()` macro is redefined in `partial_eq` tests as `asset!(left != right)`.

## Checklist

- [x] Tests are added/updated


JelteF:
Do we want to do this by default? It'll trigger twice as much code to be generated. As an option I think this makes a lot of sense, but I'm on the fence about doing this by default.

tyranron:
@JelteF I do think that it should be default behavior when proxying implementations from the underlying types, which is 90% use cases of this macro. Not being default leads to the situations where people overlook this far too often.

tyranron:
@JelteF the situations when this is not needed are actually when `fn ne()`'s body is either `false` or plain `mem::discriminant()` comparison. Thinking about it a little bit more, we can recognize these exact situations when expanding the macro. I'll add this optimization shortly.

tyranron:
@JelteF done. Now, the `ne()` method is generated only whenever `!=` comparison of the underlying types is involved.

Issue #397, opened 2024-08-12
Clippy attributes doesn't work with `Constructor`


tokarevvit:
For example when tried to apply `#[allow(clippy::too_many_arguments)]` it has no effect.
But it works for `Debug`.

```
use derive_more::Constructor;

#[allow(clippy::too_many_arguments)]
#[derive(Debug, Constructor)]
struct Test{
    a: u64,
    b: u64,
    c: u64,
    d: u64,
    e: u64,
    f: u64,
    g: u64,
    h: u64,
    i: u64
}

fn main() {
    let test = Test::new(1,2,3,4,5,6,7,8,9);

    println!("{:?}", test);
}
```

`cargo expand` output:
```
#![feature(prelude_import)]
#[prelude_import]
use std::prelude::rust_2021::*;
#[macro_use]
extern crate std;
use derive_more::Constructor;
#[allow(clippy::too_many_arguments)]
struct Test {
    a: u64,
    b: u64,
    c: u64,
    d: u64,
    e: u64,
    f: u64,
    g: u64,
    h: u64,
    i: u64,
}
#[automatically_derived]
#[allow(clippy::too_many_arguments)]
impl ::core::fmt::Debug for Test {
    #[inline]
    fn fmt(&self, f: &mut ::core::fmt::Formatter) -> ::core::fmt::Result {
        let names: &'static _ = &["a", "b", "c", "d", "e", "f", "g", "h", "i"];
        let values: &[&dyn ::core::fmt::Debug] = &[
            &self.a, &self.b, &self.c, &self.d, &self.e, &self.f, &self.g, &self.h, &&self.i,
        ];
        ::core::fmt::Formatter::debug_struct_fields_finish(f, "Test", names, values)
    }
}
#[allow(missing_docs)]
#[automatically_derived]
impl Test {
    #[inline]
    pub const fn new(
        a: u64,
        b: u64,
        c: u64,
        d: u64,
        e: u64,
        f: u64,
        g: u64,
        h: u64,
        i: u64,
    ) -> Test {
        Test {
            a: a,
            b: b,
            c: c,
            d: d,
            e: e,
            f: f,
            g: g,
            h: h,
            i: i,
        }
    }
}
fn main() {
    let test = Test::new(1, 2, 3, 4, 5, 6, 7, 8, 9);
    {
        ::std::io::_print(format_args!("{0:?}\n", test));
    };
}
```

JelteF:
[...]

PR #477, opened 2025-05-29
Add attribute too_many_arguments to the new method for Construct

Cheban1996:
Resolves https://github.com/JelteF/derive_more/issues/397

## Synopsis

Clippy emits a too_many_arguments warning for the new method generated by the Constructor derive.
Clippy attributes currently have no effect in this context.

## Solution

Add #[allow(clippy::too_many_arguments)] to the generated new method to suppress the warning without affecting other code.


tyranron (approved):
@Cheban1996 thanks for the PR and effort!

However, I kinda dislike the idea of ultimately silencing this particular lint. It could lead to the situations where some code smell is overlooked. It's better to proxy-pass `#[allow]` attributes from the type and left it to the user decision.

I've reimplemented accordingly, and also added tests.

PR #482, opened 2025-06-03
more FromStr attributes

### LINKED ISSUE #480: more `FromStr` attributes
#102 is similar, but I'm using this issue to expand and propose the design for a PR I'm writing:

the `FromStr` derive as-is still doesn't capture all use cases, so I'd like to propose some more attributes, based on serde's available attributes:

- `#[from_str(rename = "...")]`
  - available for unit struct container or enum variants
  - incompatible with `rename_all` on the same container/variant
  - changes the string value from the one derived from the container/variant ident to the string literal value
- `#[from_str(default)]`
  - only valid for exactly 1 enum variant
  - incompatible with `rename_all` and `rename` on the same variant
  - this variant is returned only when no other variant is successfully parsed
  - changes `FromStr::Error` to be `Infallible`
- `#[from_str(forward)]` (aka `#[serde(transparent)]`)
  - only valid for enum variants
    - this could also be used for allowing newtype structs
  - forwards the `FromStr` implementation to the field type
  - only allows for a single inner tuple type (eg `Foo(Inner)`)
- `#[from_str(alias = "...")]`
  - available for unit struct container or enum variants
  - allows for matches of the given string literal to result in construction of the container/variant
  - incompatible with `forward` and `default`
  - `rename_all` case conversions don't affect this value
  - `rename` is *not* required to use this attribute
- `#[from_str(skip)]`
  - only available for enum variants
  - skips *any* string from matching this variant
  - incompatible with all other attributes

dotcarmen:
closes #480 

## Synopsis

as mentioned in #480, there are some more attributes that would be nice to have in the FromStr implementation

## Solution

attributes added:
- `#[from_str(alias = "...")]` allows for additional strings to result in the variant or unit struct to be built
- `#[from_str(default)]` changes the error type to be `Infallible` and results in the annotated enum variant being constructed if no other variants are matched
- `#[from_str(forward)]` allows forwarding an enum variant's matching and field construction to the field's type
- `#[from_str(rename = "...")]` overrides the string that causes the enum variant or unit struct to be matched. that is - instead of matching based on the variant ident, the given string literal is matched
- `#[from_str(skip)]` causes the given enum variant to not be matched or constructed at all

there are various checks as necessary to ensure that attributes don't interfere with each other or are redundant:
- `default`, `forward`, and `skip` - if present, must be the only attribute on the enum variant or struct
- `rename` and `rename_all` are incompatible with each other

## Checklist


dotcarmen:
i'll leave this as a draft until i fix tests and add docs :) feedback appreciated in the meantime

i had to do a bit of a refactor, i hope i kept things tidy enough :)

impl/src/from_str.rs
```diff
@@ -249,64 +309,466 @@ impl ToTokens for FlatExpansion<'_> {
             self.self_ty.1.split_for_impl();
         let ty_name = ty.to_string();
 
-        let scrutinee_lowercased = self
-            .rename_all
-            .is_none()
-            .then(|| quote! { .to_lowercase().as_str() });
```
dotcarmen, line 255: i was able to eliminate this allocation by using `eq_ignore_ascii_case` instead :)

dotcarmen: now that i think about it, it'd be nice if `convert_case` supported checking across cases as well...

dotcarmen: maybe i can generate matchers for snake and kebab case as well? up to the reviewer :)

PR #488, opened 2025-06-27
Check release date on CI

tyranron:
I often catch myself that I forgot to update the release date in CHANGELOG once preparing a new release. So, I've added such check to CI in the projects I maintain. Would be useful here too, I guess.


tyranron:
@JelteF yes, it checks when the Git tag is pushed, so if the date is incorrect, the GitHub release won't be created, and the tag could be re-applied to the next commit fixing it.

Regarding publishing to crates.io, it's usually done via CI in my projects. But as far as I recall, you were hesitant to that, to avoid leaking potential access to other your crates from this GitHub repo. Since then, crates.io added functionality for granular token access. It could be configured like this to not allow anything more:
Screenshot 2025-06-30 at 16 52 00

If you're not hesitant to this idea anymore, I can also add publishing via CI again, and do the CHANGELOG before anything is published. Git tag, however, is the only thing to be reapplied if CI fails.

tyranron:
@JelteF thoughts?

tyranron:
@JelteF ping

tyranron:
@JelteF we even can go now with the [trusted publishing](https://blog.rust-lang.org/2025/07/11/crates-io-development-update-2025-07/#trusted-publishing) if you'd like, so there would be no need to generate a token.

PR #503, opened 2025-09-17
Support for custom error type in `TryInto` derive

### LINKED ISSUE #396: Can’t define traits on TryIntoError
I’m working with a third-party crate which defines a `ClientError` type and a `From<&'static str>` trait for that type.  Further, the crate has a generic method with the following two bounds: `T: TryInto<Foo>` and `T::Error: Into<ClientError>`.  Lastly, in my code I have an enum whose one of the variant carries `Foo` value and uses `derive_more::TryInto` to define the required `TryInto<Foo>` implementation.

With derive_more 0.99 this worked fine.  derive_more created the implementation with `&'static str` error which satisfies `Into<ClientError>` bound.

Sadly, with derive_more 1.0 the conversion error is now `TryIntoError` which cannot be converted into `ClientError` and because both of those types are third-party I cannot add the implementation.

I wonder if it would be possible to add a customisable error.  For example I could then do:

```rust
#[derive(derive_more::TryInto)]
#[try_into(error = MyError)]
pub enum Blah {
    Foo(third_party::Foo),
    Bar(third_party::Bar),
    Baz(third_party::Baz),
}

#[derive(derive_more::From)]
struct MyError(derive_more::TryIntoError<Blah>);
```

and then define all necessary conversions on `MyError`.  Even more flexible if I could define error type and constructor separately.

This actually is something I’ve suggested a while back, see https://github.com/JelteF/derive_more/issues/315, and back then I concluded that TryIntoError would work for my case.  Alas, turns out not quite.

PS. variant_names and output_type be pub?

kiendang:
Resolves #396

## Synopsis
Add support for specifying a custom error type for `TryInto` derive via a `#[try_into(error(error_ty[, error_fn]))]` attribute similarly to #494 for `FromStr`.

However there are a couple of issues that need clarifying 

1. `TryInto` derive supports `[#try_into(owned, ref, ref_mut)]` attributes which are handled by `State::with_attr_params`. To add support for the error attribute, I took the approach that requires less amount of code change: find the attributes in `input.attrs` that follow the `[#try_into(error` pattern, parse them with `attr::Error` for the custom error support. These attributes are also taken out of `input.attrs` before the `State::with_attr_params` call, leaving the rest of the `TryInto` derive code unchanged. This works. However, this leads to the error message being unhelpful in cases like this
    ```rust
    #[try_into(owned, ref, ref_mut)]
    #[try_into(asdf)] // some invalid attribute
    ^Only a single attribute is allowed
    ```
    "Only 1 `[#try_into(owned, ref, ref_mut)]` and 1 `#[try_into(error(...))]` attribute allowed" would be more helpful.

2. The PR currently works if the custom error is non-generic, *e.g* `struct CustomError(String)` (see the added test for a working example). However, in this case
    ```rust
    struct CustomError<T>(TryIntoError<T>)
    ```
    there is a problem with specifying the type for `ref` and `ref_mut`.
    ```rust
    struct CustomError<T>(TryIntoError<T>);

    impl<'a, T> From<TryIntoError<&'a T>> for CustomError<&'a T> {
        fn from(value: TryIntoError<&'a T>) -> Self {
            Self(value)
        }
    }

    #[try_into(ref)]
    #[try_into(error(CustomError<&MixedInts>))]
                                 ^this lifetime must come from the implemented type
                                 ^expected named lifetime parameter
    enum MixedInts {
        SmallInt(i32),
        NamedBigInt {
            int: i64,
        },
    }
    ```

    Another thing is suppose we want to derive for all 3 `owned`, `ref` and `ref_mut`, the custom error types might be different for each of them. The syntax can be extended to `#[try_into(error(owned(...), ref(...), ref_mut(...)))]` to accommodate this, but the issue above needs solving first.

## Checklist

### DIFF
--- a/tests/try_into.rs
+++ b/tests/try_into.rs
@@ -1,4 +1,5 @@
 #![cfg_attr(not(feature = "std"), no_std)]
+#![allow(dead_code)] // some code is tested for type checking only
 
 #[cfg(not(feature = "std"))]
 extern crate alloc;
@@ -10,7 +11,6 @@ use derive_more::TryInto;
 
 // Ensure that the `TryInto` macro is hygienic and doesn't break when `Result`
 // has been redefined.
-#[allow(dead_code)] // hygiene
 type Result = ();
 
 mod enums {


tyranron (commented):
@kiendang 
> 1. `TryInto` derive supports `[#try_into(owned, ref, ref_mut)]` attributes which are handled by `State::with_attr_params`.

> This works. However, this leads to the error message being unhelpful in cases like this

A very dirty hack could be used after parsing `State::with_attr_params()`, like: checking the returned error message and putting the better one. Covering this with test will ensure no regression is happened if the error from `State::with_attr_params()` is changed accidentally.

Not an optimal solution, of course, but does the job, until handling `owned`/`ref`/`ref_mut` is refactored with the new `utils::attr` machinery.

> 2. The PR currently works if the custom error is non-generic

I think this could easily be supported like this:
```rust
struct CustomError<T>(TryIntoError<T>);

impl<T> From<TryIntoError<T>> for CustomError<T> {
    fn from(value: TryIntoError<T>) -> Self {
        Self(value)
    }
}

impl<T> CustomError<T> {
    fn new(err: TryIntoError<T>) -> Self {
        Self(err)
    }
}

fn into_custom_err<T>(err: TryIntoError<T>) -> CustomError<T> {
    err.into()
}

#[derive(TryInto)]
#[try_into(ref)]
#[try_into(error::<T>(CustomError<T>))]
// or
#[try_into(error::<T>(CustomError<T>, CustomError<T>::new))]
// or
#[try_into(error::<T>(CustomError<T>, into_custom_err<T>))]
enum MixedInts {
    SmallInt(i32),
    NamedBigInt { int: i64 },
}
```
where `T`, of course, would be substituted with the appropriate type, like it's done for `derive_more::TryIntoError` in the expansion.

This requires a slight extension of the `attr::Error` machinery. Note, that this behavior should be controlled, in a way that `error::<T>` is not allowed for `derive(FromStr)` at all. Maybe, parametrizing `attr::Error` with behaviour marker will do the trick:
```rust
pub(crate) struct Error<const GENERICS: bool = false> {
    // ...
}

/// [`Error`] attribute with type parameters support (e.g. `error::<T>(...)`).
pub(crate) type ErrorGeneric = Error<true>;
```
And so, `derive(TryInto)` will use `attr::ErrorGeneric`, while `derive(FromStr)` will continue to use `attr::Error` which disallows using `error::<T>` syntax.

kiendang:
> A very dirty hack could be used after parsing `State::with_attr_params()`, like: checking the returned error message and putting the better one. Covering this with test will ensure no regression is happened if the error from `State::with_attr_params()` is changed accidentally.

Ah yes this would work.

> I think this could easily be supported like this:

This would work. My concern is if this syntax is too specific to this use case. I'm not sure how useful or common this use case of having a custom error parametrized on `T` is. In the original issue #396, the user just wants to have a `&'static str` error (making `output_types` and `variant_names` public member of `TryIntoError` like [Jelte suggested](https://github.com/JelteF/derive_more/issues/396#issuecomment-2332685835) would be useful here. We can add that). Another thing is the syntax is a bit confusing. Looking at

```rust
#[try_into(error::<T>(CustomError<T>, into_custom_err<T>))]
enum MixedInts {
    SmallInt(i32),
    NamedBigInt { int: i64 },
}
```

it's a bit unclear what `T` means. It's not immediately obvious that `T` refers to `MixedInts` and is related to `TryIntoError<T>`.

kiendang:
We don't even have to support that use case now, though. We can support only non-generic error type now. Extending the syntax to `#[try_into(error(owned(...), ref(...), ref_mut(...)))]` and/or `#[try_into(error::<T>(CustomError<T>))]
` is a backward-compatible change anw.

tyranron:
@kiendang 

> My concern is if this syntax is too specific to this use case. I'm not sure how useful or common this use case of having a custom error parametrized on `T` is.

Well, the point is:
- If you don't use define your error like this `struct CustomError<T>(TryIntoError<T>);`, you don't need to use this turbofish `::<T>` parametrization and just `#[try_into(error(CustomError))]` does the job.
- If, however, you do want to preserve the original value in the error, then you should specify the parametrization for the macro to understand which parameter should be substituted. Usually `for<T>` is used for such cases, like `#[try_into(error(for<T> CustomError<T>))]`. However, we have to attribute arguments here, and the custom conversion function will need to specify parametrization too very much likely, which requires repeating that `for<T>` twice (e.g. `#[try_into(error(for<T> CustomError<T>, for<T> CustomError<T>::new))]`). So, I thought turbofish would be more ergonomic in this case, defining the parameter for all arguments, like Rust does in functions. Maybe, instead turbofish `::<T>`, just `<T>` would be enough: `#[try_into(error<T>(CustomError<T>, CustomError<T>::new))]`? I'm open to better syntactic suggestions, of course.

So, overall, you don't use weird syntax if you don't have to.

> it's a bit unclear what `T` means. It's not immediately obvious that `T` refers to `MixedInts` and is related to `TryIntoError<T>`.

That's what documentation is for!

> We don't even have to support that use case now, though.

I'm OK with not supporting it at all! Only giving suggestions how we could solve this if we want to!
Though, this caveat should be explicitly stated in docs.

kiendang:
> I'm OK with not supporting it at all! Only giving suggestions how we could solve this if we want to!
Though, this caveat should be explicitly stated in docs.

Yup I agree. Let's support just the simple non-generic case for this PR.

Still, I gave much thought to the syntax for the generic case so I'll just put them down here anw.
> If, however, you do want to preserve the original value in the error, then you should specify the parametrization for the macro to understand which parameter should be substituted.

I agree with this. The most appropriate syntax for this case would probably look like what you proposed, `struct CustomError<T>(TryIntoError<T>);`.

1. A generic parameter `T`, without any type constraint, usually would mean any type could substitute it. However `T` does not mean just any type here. It has to be one of `MixedInt` `&MixedInt` or `&mut MixedInt` where appropriate, which is not shown in the syntax. There is documentation of course, just that (a) this is a bit out of the normal convention and (b) if I understand correctly, code for automatic derivation should aim to be obvious to read.
2. Usually (not always) in a generic function/trait with a generic parameter `T`, users would expect to look into the body of the function/trait to see where `T` is used.

> I'm open to better syntactic suggestions, of course.

Originally, I thought of something like `#[try_into(error(owned(MixedInt), ref(&MixedInt), ref_mut(&mut MixedInt)))]`, which is a bit verbose, but obvious. However, this doesn't work. It's not even correct. Besides the lifetime annotation issue, in this example

```rust
enum MixedInts {
    SmallInt(i32),
    NamedBigInt { int: i64 },
    Unit
}
```

the error type for the `Unit` case for `ref` is `TryIntoError<MixedInt>`, not `TryIntoError<&MixedInt>`! In comparison, your syntax `struct CustomError<T>(TryIntoError<T>);` with `T` means *one of `MixedInt` `&MixedInt` or `&mut MixedInt` where appropriate* is not only more concise, but also correct.

Maybe `struct CustomError<T: MixedInt | &MixedInt | &mut MixedInt>(TryIntoError<_>);`? I might just make it worse though.

kiendang:
@tyranron sorry for the delay.

The failed CI seems unrelated. I submitted #508 to fix.

I've added to doc and adjusted the error message as suggested

> A very dirty hack could be used after parsing `State::with_attr_params()`, like: checking the returned error message and putting the better one. Covering this with test will ensure no regression is happened if the error from `State::with_attr_params()` is changed accidentally.

I want to adjust the message for this error as well, but I can't seem to be able to trigger it. Looks like both `#[tryinto]` and `#[tryinto()]` are allowed?
https://github.com/JelteF/derive_more/blob/e0a0713f93ec38d777fbcb789ea1e41dffc415b6/impl/src/utils.rs#L858

For the surplus attribute case I changed the message to ```Only a single attribute, plus a optional `#[try_into(error(...))]`, is allowed```, but obviously the optional `#[try_into(error(...))]` is only applicable to the top level attribute, not the nested one. Should I adjust it to ```Only a single attribute, plus a optional `#[try_into(error(...))] if it is a top level attribute`, is allowed```? Is there a more precise/formal term than top level attribute?

tyranron:
@kiendang 

> Looks like both `#[tryinto]` and `#[tryinto()]` are allowed?

Maybe... it's quite entangled. Maybe an additional check is required in such a case?

> Should I adjust it to `` Only a single attribute, plus a optional `#[try_into(error(...))] if it is a top level attribute`, is allowed ``? Is there a more precise/formal term than top level attribute?

"Top level" is a normal term for that. However, I think it makes sense to emit just different errors for top-level and variant-level.

kiendang:
> "Top level" is a normal term for that. However, I think it makes sense to emit just different errors for top-level and variant-level.

I agree, but for just adjusting the error message emitted from `State::with_attr_params()` we can't really differentiate. I think we can do that more easily once attributes are refactored with the `utils::attr` machinery as you said?

PR #508, opened 2025-10-20
Fix ci

kiendang:
## Synopsis

Fix a couple of CI failures due to cargo/rustc updates
- replace `doc_auto_cfg` with `doc_cfg` rust-lang/rust#138907
- update `as_ref`, `as_mut` compile error message

## Checklist


kiendang:
Hmm looks like the error message changes in nightly but not in stable yet hence updating it fails stable.

tyranron (approved):
@kiendang thank you for taking care of it!

> Hmm I cannot reproduce locally. 

Don't bother. This flakiness is here for years now. Nothing we can do about it, though.

PR #532, opened 2026-01-07
Add Hash derive

fredszaq:
## Synopsis

This PR adds a new derive for the `Hash` trait. This does mostly the same as the derive bundled with rust std but with a few niceties :
 - no overcontrain of generic types
 - possibility to ignore fields with `#[hash(skip)]` (also honors `#[eq(skip)]` and `#[partial_eq(skip)]` to ensure trait expectations) 
 - possibility to use custom hash function on fields with `#[hash(with(my_custom_hash_function)]`

## Solution

I used the `PartialEq` derive as an example to implement this, didn't have to do exotic things to get things to work

## Checklist


tyranron (changes requested):
@fredszaq that's a solid work!

~~One major concern should be addressed before merging this.~~ Otherwise, the PR makes a very good impression!

Thank you very much for contributing this and being pedantic to the implementation!

UPD: The concern [appeared](https://github.com/JelteF/derive_more/pull/532#discussion_r2729417618) to be not major and non-concern at all. Then, will try to merge it the next few days!

fredszaq:
Hey @tyranron, pinging back on this, anything I can do to help you get this merged ?

tyranron:
@fredszaq 

> anything I can do to help you get this merged ?

Nah... I would like to, but don't see how you could possibly decrease the load on my current job for me to get back on this 😅

Anyway, thanks for asking! I will reall try to merge this on these weekends.

fredszaq:
No pressure^^ do not hesitate if you have any questions or want me to make changes !

.gitignore
```diff
@@ -16,3 +16,6 @@ gh-pages
 # git files
 .swp
 /tags
+
+# RustRover/Other jetbrains IDE files
+.idea/
```
tyranron: @fredszaq this better be configured globally on your side.

impl/src/utils.rs
```diff
@@ -2393,6 +2404,35 @@ pub(crate) mod attr {
...
     }
+
+    #[cfg(feature = "hash")]
+    mod with {
+        use crate::utils::attr::ParseMultiple;
+        use syn::parenthesized;
+        use syn::parse::{Parse, ParseStream};
+
+        pub struct With {
+            pub path: syn::Path,
```
tyranron: @fredszaq similarly to `#[from_str(error(..))]` it would be nice to support `syn::ExprCall` and `syn::ExprClosure` here too. But let's do this in a separate PR.

tyranron (approved):
@fredszaq thanks for the patience!

impl/src/utils.rs
fredszaq: done in #548

PR #533, opened 2026-01-09
Add Clone and Copy derives

### LINKED ISSUE #439: Add support for deriving Clone/Copy and other standard traits without generic type parameter transitiveness
The built-in `#[derive(Copy, Clone)]` and others in Rust have this rule that all generic type parameters must also implement `Copy` or `Clone`. Perhaps the logic relies on this to ensure that all fields really can be copied/cloned/compared?

However, this is not always desirable. For example, some generic type parameter might simply be used by reference:

```rust
trait Bar {}

#[derive(Copy, Clone)]
struct Foo<B: Bar> {
    bar: &'static B
}
```

While the reference `bar` is perfectly copyable, Rust will complain here and say that `Copy` and `Clone` cannot be derived because `Bar` does not have them as supertraits: [Rust Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=ac5e33bee9b3c1a5ffa555044f60a3ae).

```
error[E0599]: the method `clone` exists for struct `Foo<BarImpl>`, but its trait bounds were not satisfied
  --> src\main.rs:14:20
   |
4  | struct Foo<B: Bar> {
   | ------------------ method `clone` not found for this struct because it doesn't satisfy `Foo<BarImpl>: Clone`
...
8  | struct BarImpl;
   | -------------- doesn't satisfy `BarImpl: Clone`
...
14 |     let foo2 = foo.clone();
   |                    ^^^^^ method cannot be called on `Foo<BarImpl>` due to unsatisfied trait bounds
   |
note: trait bound `BarImpl: Clone` was not satisfied
  --> src\main.rs:3:16
   |
3  | #[derive(Copy, Clone)]
   |                ^^^^^ unsatisfied trait bound introduced in this `derive` macro
help: consider annotating `BarImpl` with `#[derive(Clone)]`
   |
8  + #[derive(Clone)]
9  | struct BarImpl;
```

However, this check is excessively restrictive. Manually implementing Copy and Clone works fine because the type `B` is only used as a reference.

```rust
impl<B: Bar> Copy for Foo<B> {}
impl<B: Bar> Clone for Foo<B> {
    fn clone(&self) -> Self {
        Self { bar: self.bar.clone() }
    }
}
```

The above compiles successfully and copies/clones the reference regardless of whether `B` itself impelments any of these traits.

```rust
trait Bar {}

#[derive(CopyMore, CloneMore)]
struct Foo<B: Bar> {
    bar: &'static B
}
```

It would be very useful to have "explicit" variations of the built-in derive-macros for such situations, so we could `#[derive(CopyMore, CloneMore, EqMore)]` and just have an implementation generated that goes field by field without looking at the types and leaves it up to the compiler to detect actual field by field compatibility in the generated code.

fredszaq:
Resolves #439 

## Synopsis

This adds 2 new derives for `Clone` and `Copy`. Those won't put any additionnal contraints on generic types by default, but provide `#[clone(bound(...))]` and `#[copy(bound(...))]` attributes if you need a fine tune the bounds

## Solution

I moved the `Bounds` attribute from the `fmt` module to `utils` and reused it for both Derives 

## Checklist


impl/src/utils.rs
```diff
@@ -1702,6 +1720,87 @@ pub(crate) mod attr {
...
+
+                content
+                    .parse_terminated(syn::WherePredicate::parse, token::Comma)
+                    .map(Self)
+            }
+        }
+
+        impl Bounds {
+            /// Errors in case legacy syntax is encountered: `bound = "..."`.
+            fn check_legacy_fmt(input: ParseStream<'_>) -> syn::Result<()> {
```
tyranron: @fredszaq I think after two major releases and few years passed, we may remove checks for legacy syntax.

impl/src/copy.rs
```diff
@@ -0,0 +1,24 @@
...
+    let generics = &input.generics;
+    let (impl_generics, ty_generics, where_clause) = generics.split_for_impl();
+
+    let ident = quote::format_ident!("copy");
+    let real_where_clause = attr::Bounds::parse_attrs(&input.attrs, &ident)?
+        .map(|clause| {
+            let clause = clause.item.0;
+            quote! { where #clause }
+        })
+        .or(where_clause.map(ToTokens::to_token_stream));
```
tyranron: @fredszaq see... there are two possible modes of applying custom `where` bounds:
1. Replacing: just use the provided ones (like here).
2. Additive: intermix the provided ones together with inferred ones.

For now, some macros use the first one, other the second one, reusing the same syntax for it. I don't like this situation. It's hard to remember or understand the concrete behavior for the concrete macro and requires either peeking into the macro expansion or reading the documentation every time.

I think these two modes should be clearly separated on the syntax level. For example:
```rust
#[bound(T: Trait)] // additive mode
#[bound(overwrite(T: Trait))] // replacing mode
```

I'm not sure of the concrete syntax, though, and would appreciate the suggestions.

tyranron (changes requested):

impl/src/utils.rs
fredszaq: I removed that ! I updated the compile fail tests to the new error message, should I just remove them ?

impl/src/copy.rs
fredszaq: @tyranron I changed the code to actually extend the where clause we get from the struct definition instead of completely replacing it, this actually makes way more sense. 

This makes `Copy` and `Clone` derives behave more in the additive fashion, even though they don't do clever type inference like the `Display`-family ones. 

Do you think that is enough to have both uses cases use the same syntax ? (I didn't see other usages of `where` bounds in the lib)

If we wan't to actually support both types of bounds settings I guess the ideal would be 

```rust
#[extra_bound(T: Trait)] // additive mode
#[bound(T: Trait)] // replacing mode
```
but this means breaking the behavior for every user of the feature, so probably not advised, maybe
```rust
#[exact_bound(T: Trait)
```
for the replacing mode could be an option ? it feels a bit clunky thought

fredszaq:
Hey @tyranron! 

A gentle reminder about this PR and #535 that are awaiting for your review! 

Have a great day :)

PR #535, opened 2026-01-13
Add support for custom equality functions in Eq/PartialEq

fredszaq:
## Synopsis

This adds the possibility to use custom functions to perform the equality check in `PartialEq`/`Eq` derives. This is quite useful when wrapping types without a `PartialEq`/`Eq` impl

## Solution

This builds on #532 and reuse the `FieldAttributes` for `Hash`, I moved the struct to `utils::attr` and renamed it `WithOrSkip`.

I removed fields types that have a `with` attribute from the automatically added `where Type: Eq` bound automatically generated by the `Eq` derive. This seems like the thing to do as if you're using a custom function it most probably means the field doesn't implement `Eq`.  

While this is not part of the roadmap for PartialEq, defined in #163 this is something that will be quite useful for people migrating form `derivative` as this was something supported by this crate

## Checklist

### DIFF
--- a/impl/src/cmp/partial_eq.rs
+++ b/impl/src/cmp/partial_eq.rs
@@ -36,24 +36,28 @@ pub fn expand(input: &syn::DeriveInput, _: &'static str) -> syn::Result<TokenStr
                 let mut alternate_eq_functions =
                     FieldsWithAlternateEqFunction::default();
                 'fields: for (n, field) in data.fields.iter().enumerate() {
-                    for attr_name in [&attr_name, &secondary_attr_name] {
-                        match attr::WithOrSkip::parse_attrs(&field.attrs, attr_name)? {
-                            Some(Spanning {
-                                item: attr::WithOrSkip::Skip,
-                                ..
-                            }) => {
-                                _ = skipped_fields.insert(n);
-                                continue 'fields;
-                            }
-                            Some(Spanning {
-                                item: attr::WithOrSkip::With(with),
-                                ..
-                            }) => {
-                                alternate_eq_functions.insert(n, with.func);
-                                continue 'fields;
-                            }
-                            None => {}
+                    match attr::WithOrSkip::parse_attrs(&field.attrs, &attr_name)? {
+                        Some(Spanning {
+                            item: attr::WithOrSkip::Skip,
+                            ..
+                        }) => {
+                            _ = skipped_fields.insert(n);
+                            continue 'fields;
                         }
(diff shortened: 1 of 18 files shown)


fredszaq:
@tyranron rebased and removed the draft status on this one as #532 is now merged. Feel free to review when you have time :)

tests/eq.rs
```diff
@@ -162,6 +162,64 @@ mod structs {
...
+            }
+
+            struct NotPartialEq(i32);
+
+            #[test]
+            fn single_field() {
+                #[derive(Eq, PartialEq)]
+                struct Foo(#[partial_eq(with(eq_special))] NotPartialEq);
+                #[derive(Eq, PartialEq)]
+                struct Bar(#[eq(with(eq_special))] NotPartialEq);
```
tyranron: @fredszaq I don't really like the idea of having `#[eq(with(...))]` attribute. For `skip` it has sense, since skipping for `Eq` and `PartialEq` is essentially the same thing. `with(...)`, however, has only sense for the `PartialEq`. Of course, `derive(Eq)` should account for the `#[partial_eq(with(...))]` presence. But having the `#[eq(with(...))]` attribute confuses me a lot, implying that `Eq` trait has a method, which is not the case.

impl/src/cmp/partial_eq.rs
```diff
@@ -84,6 +124,10 @@ pub fn expand(input: &syn::DeriveInput, _: &'static str) -> syn::Result<TokenStr
 /// Indices of [`syn::Field`]s marked with an [`attr::Skip`].
 type SkippedFields = HashSet<usize>;
 
+/// Mapping from [`syn::Field`] marked with an [`attr::With`] to the [`syn::Path`] of the alternate
+/// eq function.
+type FieldsWithAlternateEqFunction = HashMap<usize, syn::Path>;
```
tyranron: @fredszaq let's stick to the same terminology everywhere and just call it "custom function".

tyranron (changes requested):

tests/eq.rs
fredszaq: Fixed. I also added an error if you have `#[derive(Hash)]` with `partial_eq(with)` and no `hash(skip/with)`

impl/src/cmp/partial_eq.rs
fredszaq: done, changed also on the hash derive that was also using the alternate terminology

PR #538, opened 2026-01-21
chore: replace `unicode-xid` with `unicode-ident`

larseggert:
## Synopsis

The `unicode-ident` crate is the same dependency already used by `proc-macro2` and `syn`, reducing the total number of dependencies. It's also supposedly more optimized.

### DIFF
--- a/impl/Cargo.toml
+++ b/impl/Cargo.toml
@@ -27,7 +27,7 @@ proc-macro2 = "1.0"
 quote = "1.0"
 syn = "2.0.45"
 convert_case = { version = "0.10", optional = true }
-unicode-xid = { version = "0.2.2", optional = true }
+unicode-ident = { version = "1.0", optional = true }
 
 [build-dependencies]
 rustc_version = "0.4"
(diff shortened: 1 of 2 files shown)


tyranron (approved):
@larseggert originally, [`unicode-xid` was used because it was used by `syn`](https://github.com/JelteF/derive_more/pull/181#issue-1107766100). Since then, [`syn` migration to `unicode-ident`](https://github.com/dtolnay/syn/commit/be107389d4086970a4ef477aab7a80e36951c0ac) went unnoticed.

Thank you for noticing and making the effort to fix!

larseggert:
Any chance a new release will happen soon? I can't use the crate until this PR has landed in a release.

tyranron:
@larseggert I think it will take some time to produce the new one. The progress could be tracked in the milestone: https://github.com/JelteF/derive_more/milestone/7

larseggert:
Rechecking if a new release with this could be cut? The milestone above seems very far off still.

PR #560, opened 2026-08-12
Fix Constructor derive emitting redundant_field_names lint

akundaz:
The `Constructor` derive for named-field structs was generating:

```rust
Self { field: field }
```

instead of the field init shorthand:

```rust
Self { field }
```

This triggers `clippy::redundant_field_names` on nightly Clippy when `-D warnings` is set, breaking crates that use `#[deny(clippy::all)]` or similar.

**Fix:** use `#(#field_names),*` in the struct literal rather than `#(#field_names: #vars),*`. Since the parameter names already match the field names, the shorthand is always valid here. This also removes a redundant variable alias and clone in `struct_body`.


akundaz:
The CI failure (`compile_fail`) seems to be pre-existing and unrelated to this change. I can reproduce it locally on master without my commit applied — 4 of 136 compile_fail tests fail there too.

PR #561, opened 2026-08-19
Add `IsVariantAnd` derive

### LINKED ISSUE #365: `IsVariantAnd` for enums
Just like `Option` has `is_some_and()`, could we have a `is_variant_and()` derive?

I may try making a pull request, but I don't think I know enough to actually do everything

ChrisJr404:
Closes #365.

This adds an `IsVariantAnd` derive: for each enum variant `foo` it generates an `is_foo_and(&self, f) -> bool` method that returns `true` only when the value is that variant and the given closure returns `true` for its contents. It's the `Option::is_some_and` pattern generalised to any enum, which is handy for matching on a variant and checking something about its payload in a single expression instead of a `matches!` or a `match` arm.

I modelled the API on the existing `IsVariant` and `Unwrap` derives:

- The method takes `&self` (like `IsVariant`), so it's non-consuming and works on borrowed enums.
- The closure receives the variant's fields by reference as a tuple, in declaration order, the same way `Unwrap` returns them. A single-field variant collapses to a plain reference, so `Just(T)` gives `is_just_and(|x: &T| ...)`, exactly mirroring `Option::is_some_and`.
- Field-less variants (unit, empty tuple `()` or empty record `{}`) take a `FnOnce() -> bool` closure instead.
- `#[is_variant_and(ignore)]` skips a variant, consistent with `#[is_variant(ignore)]`.

It's behind a new `is_variant_and` feature flag (added to `full`), with docs in `impl/doc/is_variant_and.md`, a CHANGELOG entry, and integration tests in `tests/is_variant_and.rs` covering tuple/struct/unit variants, references, generics with bounds, `ignore`, a single-variant enum, and the `nightly`/`deprecated` cases the `IsVariant` tests exercise.

One thing worth a look during review: I went with references-in-a-tuple for the closure argument rather than consuming `self` like `Option::is_some_and` does, since a consuming derive felt surprising here and `IsVariant` already established `&self`. Happy to change the shape if you'd prefer something else.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Add `Hash` derive similar to `std`'s one, but considering generics correctly,
   and supporting custom hash functions per field or skipping fields.
   ([#532](https://github.com/JelteF/derive_more/pull/532))
+- Add `IsVariantAnd` derive, generating an `is_foo_and()` method for each enum
+  variant `foo` that also applies a closure to the variant's fields, similar to
+  `Option::is_some_and`.
+  ([#561](https://github.com/JelteF/derive_more/pull/561),
+   [#365](https://github.com/JelteF/derive_more/issues/365))
 
 ### Fixed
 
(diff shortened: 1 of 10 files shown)


ChrisJr404:
The failing stable check was the eq/non_eq_field compile_fail snapshot, not this derive. Newer stable rustc dropped the 'see issue #48214' help line, so I reblessed that .stderr to match. Passes locally now with --features full.

